PR #2, opened 2017-05-27
Initiate the project

minggangw:
This patch implements the basic framework of binding the RCL library
to the JavaScript interface, meanwhile brings out the timer object and
with an example of it.

The project is using eslint (JavaScript) and cpplint (c/c++) to check
the static syntax. Hopefully, unit tests will be added later. And the
buildbot is running on travis-ci with docker srevice.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,46 @@
+env:
+  node: true
+  es6: true
+
+globals:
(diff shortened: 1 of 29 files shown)


jihoonl:
:+1:

PR #4, opened 2017-06-12
[rosidlparser]Wrap the rosidl_parser into JavaScript

minggangw:
This patch simply realized to wrap the rosidl_parser and exposed the
final result as JavaScript object.

Calling the rosidl_parser in node.js to get the python object
as result, then wrapped the result to a JSON object and serialized it
to a string. The string will be deserialized to JSON object in
the JavaScript at least.

There are two examples to illustrate the usage. One shows how to parse
the message file (.msg), the other shows how to parse the service file
(.srv).


minggangw:
@kenny-y PTAL thanks!

kenny-y:
Considering the current phase of development... I'm OK with whatever code that works :)

PR #6, opened 2017-06-12
Update the Node.js version to latest LTS Boron, v6.11.0

minggangw:


minggangw:
Just update the Node.js version, merging.

PR #8, opened 2017-06-14
Fix the coding convention problems

minggangw:
This patch fixed some prombles which broke the coding conventions to
make it consistent. More, to use the NAN macro to define the binding
functions to make the code simple.


minggangw:
Merge this pr as it only fix the convention problems.

PR #10, opened 2017-06-15
Generate all message class(es) when initializing rclnodejs

kenny-y:
TODO: change this default behavior by different policies.


minggangw (approved):
I notice that there are some meaningless comments and invalid endings, please correct them when merging it, other part looks good to me, thanks!

PR #11, opened 2017-06-16
Generator unit test

kenny-y:
Adding the first 2 test cases to cover: (1) the message class generation of all ros messages (2) the usage of the generated message class of std_msgs/msg/String.msg


minggangw (approved):
LGTM, thanks for the patch:) and merge.

Issue #12, opened 2017-06-16
Check arguments in JavaScript before passing them to C++ addon


kenny-y:
Do NOT to do such checks in C++ (or only do the minimum in C++)

minggangw:
I'm going to take this issue.

minggangw:
This issue has been fixed by 673d9d25ee3e9

PR #18, opened 2017-06-20
Check the arguments on JavaScript side

minggangw:
Now we validated the argrments, which are passed from JavaScript, in C++ addon.
This escalated the cost, and as the binding methoths will be called by our
JS interfaces only, it's better to do the verification in JavaScript
instead.

Fix issue #12

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,14 +31,16 @@ let rcl = {
   _nodes: [],
 
   createNode(nodeName, namespace = '') {
-    let handle = rclnodejs.createNode(nodeName, namespace);
+    if (typeof (nodeName) !== 'string' || typeof (namespace) !== 'string') {
+      throw new TypeError('Invalid argument.');
+    }
 
+    let handle = rclnodejs.createNode(nodeName, namespace);
     let node =  new rclnodejs.ShadowNode();
+
     node.init();
     node._handle = handle;
-
     this._nodes.push(node);
-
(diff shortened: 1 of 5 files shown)


minggangw:
@kenny-y PTAL, thanks!

index.js
```diff
@@ -31,14 +31,16 @@ let rcl = {
   _nodes: [],
 
   createNode(nodeName, namespace = '') {
-    let handle = rclnodejs.createNode(nodeName, namespace);
+    if (typeof (nodeName) !== 'string' || typeof (namespace) !== 'string') {
+      throw new Error('Wrong type of argments.');
```
kenny-y: Consider using TypeError: "The TypeError object represents an error when a value is not of the expected type" from:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

kenny-y: Ref: https://github.com/RobotWebTools/rclnodejs/blob/develop/rosidl_gen/templates/message.dot#L117

kenny-y (approved):
Apart from the TypeError change request, the rest looks good. I'll make the check mark green in advance.

index.js
minggangw: Yes, TypeError should be used, will correct and update the pr, thanks!

PR #23, opened 2017-06-27
Adding basic array support and test cases for ROS messages

kenny-y:
Unresolved issues will be addressed later


minggangw (approved):
Will commit after pr#24 merged

PR #25, opened 2017-06-28
Fixing a crash of non-zero filling Node.js Buffer of old Node.js vers…

kenny-y:
…ions.

Solution: call rosidl_generator_c__String__init() to init the structure, because in rosidl_generator_c__String__assign(), the ->data will be free()


minggangw (approved):
LGTM, waiting for the ci then merge, thanks!

PR #26, opened 2017-06-28
[subscription]Ensure to pass the correct variable when receiving messages

minggangw:
The current implementation used wrong variable to pass, when it called the
callback function.

This patch fixed the problem and made other small changes to make the
code reasonable.

### DIFF
--- a/lib/subscription.js
+++ b/lib/subscription.js
@@ -18,31 +18,34 @@ const rclnodejs = require('bindings')('rclnodejs');
 const generator = require('../rosidl_gen/generator.js');
 
 class Subscription {
-  constructor(handle, messageType, message, topic, callback) {
+  constructor(handle, messageType, topic, callback) {
     this._handle = handle;
-    this._messageType = messageType,
     this._topic = topic;
     this._callback = callback;
-    this._Message = message;
-    this._rosMsg = new message();
+    this._Message = messageType.Message;
+    this._rosMsg = new messageType.Message();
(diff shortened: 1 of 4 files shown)


minggangw:
@kenny-y PTAL, thanks!

lib/subscription.js
```diff
@@ -18,31 +18,34 @@ const rclnodejs = require('bindings')('rclnodejs');
...
 class Subscription {
-  constructor(handle, messageType, message, topic, callback) {
+  constructor(handle, messageType, topic, callback) {
     this._handle = handle;
-    this._messageType = messageType,
     this._topic = topic;
     this._callback = callback;
-    this._Message = message;
-    this._rosMsg = new message();
+    this._Message = messageType.Message;
```
kenny-y, line 25: If we want to carry `Message` class via a member of `msgType`, we should change `generator` behavior, other than requesting client code to do this (e.g. [`subscription-example.js:28`](https://github.com/RobotWebTools/rclnodejs/pull/26/files#diff-2d202f8819d7688fdc7ca04e7bc3f6eeR28)). Or we can use `getMessageClass` function of generator to get the message class by passing `msgType` as argument.

kenny-y (approved):

lib/subscription.js
minggangw: That's right, I will make this change when the message generator make the corresponding change.

PR #32, opened 2017-07-03
Resturcture the codebase

minggangw:
After being restructured, the root will consist following folders:

root
├── example
├── lib (JavaScript class file)
├── rosidl_gen (message generator for ROS message)
├── rosidl_parser (parser for .msg/.srv file)
├── scripts
├── src (C/C++ code, including the third party)
└── test (test files)


minggangw:
@kenny-y PTAL, thanks!

kenny-y (approved):

PR #37, opened 2017-07-07
Add function to create rosidl_service_type_support_t*

minggangw:
When we are creating client/service in ROS2 through rcl library, we must
get variable of type rosidl_service_type_support_t* to initialize them. This
patch adds the interface, which can get rosidl_service_type_support_t* by
the service package name and service name.


minggangw:
@kenny-y PTAL, thanks!

kenny-y (approved):

Issue #27, opened 2017-06-28
Let the constructor of generated message class to be as a property of messageType


minggangw:
Through the current implementation, we have to get two variables (message type & message constructor) passed to the publisher/subscription, like
```
const messageType = message.getMessageType('std_msgs', 'msg', 'String');
const StringMessage = message.getMessageClass(messageType);
let msg = new StringMessage();
```
Here, I'd like to assign the constructor of the message as a property of message type (named as the message type's name), thus we can deduct the second line and pass only one parameter in most cases. The final code looks like:
```
const std_msgs = message.getMessageType('std_msgs', 'msg', 'String');
let msg = new std_msgs.String();
```
Also, the sub folder as a parameter should be removed, as the caller may not be aware of this conception. The package name (like `std_msgs`) and the message type (like `String`) are enough to locate the .msg/.srv files.

minggangw:
Fixed by 7399fee74

PR #38, opened 2017-07-12
[rosidl_gen]Change the interface of generator

minggangw:
This patch made the following changes:
1. Change the path of the .js files generated from the .msg/.srv files,
the location is ./generated/[packageName]
2. Refactor the way of getting the interface class and passing it to the
subscription/publisher.

Fix issue #27

### DIFF
--- a/example/publisher-example.js
+++ b/example/publisher-example.js
@@ -18,16 +18,15 @@
 'use strict';
 
 const rclnodejs = require('../index.js');
-const {message} = rclnodejs;
 
 rclnodejs.init().then(() => {
   const node = rclnodejs.createNode('publisher_example_node');
 
-  const messageType = message.getMessageType('std_msgs', 'msg', 'String');
-  const publisher = node.createPublisher(messageType, 'topic');
-
-  const StringMessage = message.getMessageClass(messageType);
-  var msg = new StringMessage();
+  /* eslint-disable */
+  let std_msgs = rclnodejs.require('std_msgs');
+  const publisher = node.createPublisher(std_msgs.String, 'topic');
+  let msg = new std_msgs.String();
+  /* eslint-enable */
 
   let counter = 0;
   setInterval(function() {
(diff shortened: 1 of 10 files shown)


lib/publisher.js
```diff
@@ -45,31 +47,22 @@ class Publisher {
...
-    if (message instanceof Buffer) {
-      msg = message;
-    } else if (typeof message.getBuffer === 'function') {
-      msg = message.getBuffer();
-    }  // TODO(kenny): check messageType and convert raw values to a ros message
-
-    if (msg) {
-      rclnodejs.publishMessage(this._handle, msg);
+    let rawRosMessage = this._typeClass.getRefBuffer(message);
+    if (rawRosMessage) {
```
kenny-y: I believe we still need to add an 'else if' branch to translate a plain JavaScript value/object to message if possible, this would be done by leverage the _typeClass.import() function (TODO: define such a function in dot template)
After it has been done, we can publish a string message using
```
publisher.publish('plain string');
```
or publish a 3D Point compound message using
```
publisher.publish({x: 1, y: 2, z: 3});
publisher.publish(1, 2, 3);
```

example/publisher-example.js
```diff
@@ -18,24 +18,23 @@
...
 rclnodejs.init().then(() => {
   const node = rclnodejs.createNode('publisher_example_node');
 
-  const messageType = message.getMessageType('std_msgs', 'msg', 'String');
-  const publisher = node.createPublisher(messageType, 'topic');
-
-  const StringMessage = message.getMessageClass(messageType);
-  var msg = new StringMessage();
+  /* eslint-disable */
+  let std_msgs = rclnodejs.require('std_msgs');
```
kenny-y, line 26: This is a good improvement.

kenny-y (approved):

lib/publisher.js
minggangw: It does improve the flexibility by supporting something like:
`publisher.publish('plain string')` or `publisher.publish({x: 1, y: 2, z: 3})`

Good suggestion! I will mark this as a TODO, so we can implement it later.

PR #40, opened 2017-07-20
Support to generate JavaScript struct for .srv file

minggangw:
This patch added the implementation of generating the JavaScript structs
from template, which is used by dot. Besides, decouple some functions
from generator, align with the principle of encapsulation.


minggangw:
@kenny-y PTAL, thanks!

kenny-y (approved):
Approve as long as it works.

PR #41, opened 2017-07-27
Implement the client/service feature in ROS2.0

minggangw:
This patch mainly implemented the client/service feature, the detailed
includes:

1.The JavaScript interfaces definition.
2.The binding methods bound to the RCL library.
3.Two examples: usage of client, the other is for service.
4.Some naming changes.


minggangw:
@kenny-y PTAL thanks!

kenny-y (approved):
I'm OK with this PR as long as it works. I think it's time to add some test suites for those pub/sub srv/client pairs

PR #42, opened 2017-07-31
[JSDoc]Add comments to create HTML doc generated by JSDoc

minggangw:
Please run:
  npm install -g jsdoc
to install the JSDoc globally, then run:
  jsdoc index.js lib/
The document will be put under folder 'out'.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,10 +29,20 @@ function inherits(target, source) {
 
 inherits(rclnodejs.ShadowNode, Node);
 
+/**
+ * A module that exposes the rclnodejs interfaces.
+ * @exports rclnodejs
+ */
 let rcl = {
   _initialized: false,
   _nodes: [],
 
+  /**
+   * Create a node.
+   * @param {string} nodeName - The name used to register in ROS.
+   * @param {string} namespace - The namespace used in ROS, default is an empty string.
+   * @return {Node} The instance of Node.
+   */
   createNode(nodeName, namespace = '') {
     if (typeof (nodeName) !== 'string' || typeof (namespace) !== 'string') {
       throw new TypeError('Invalid argument.');
@@ -47,6 +57,11 @@ let rcl = {
     return node;
   },
 
+  /**
+  * Init the module.
+  * @param {array} args - The command line arguments to pass to rcl.
+  * @return {Promise<undefined>} A Promise.
+  */
   init(...args) {
     return new Promise((resolve, reject) => {
       if (!this._initialized) {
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -47,6 +57,11 @@ let rcl = {
     return node;
   },
 
+  /**
+  * Init the module.
+  * @param {array} args - The command line arguments to pass to rcl.
+  * @return {Promise} A Promise.
```
kenny-y: info: `Promise<Type>` is supported here :)

index.js
```diff
@@ -65,6 +80,11 @@ let rcl = {
     });
   },
 
+  /**
+   * Start to spin the node, which triggers the event loop to start to check the incoming events.
+   * @param {Node} node - The node to be spun.
+   * @return {void}
```
kenny-y: void -> undefined?

lib/client.js
```diff
@@ -35,22 +48,45 @@ class Client {
     this._callback = callback;
   }
 
+  /**
+   * Get the handle of this client.
+   * @return {RclHandle} The handle.
```
kenny-y: The return type won't be listed in generated doc. Recommending `@property` syntax

lib/publisher.js
```diff
@@ -17,37 +17,41 @@
...
-   * @param {Object} typeClass - The class of the topic message.
+   * Create a publisher
+   * @param {RclHandle} handle - A RclHandle object of publisher.
+   * @param {RclHandle} nodeHandle - A RclHandle object of node, to which the publisher belongs.
+   * @param {function} typeClass - The class representing the message.
    * @param {string} topic - The name of the topic.
    */
-  constructor(publisherHandle, nodeHandle, typeClass, topic) {
-    this._handle = publisherHandle;
+  constructor(handle, nodeHandle, typeClass, topic) {
```
kenny-y: Is this constructor invisible to users? I'm assuming `node.createPublisher()` is the only interface we expect users to know

kenny-y (approved):

kenny-y:
This is a good start, we're gonna have to revise the document several times anyway, the issues can be fixed later or before the merge... I'm OK with either way.

index.js
minggangw: **undefined** is more suitable for JavaScript.

lib/publisher.js
minggangw: No, this constructor will not be exposed, so I'd better remove these comments for constructor.

index.js
minggangw: Oh, got this

PR #43, opened 2017-08-01
Adding customized doc templates for doc

kenny-y:
Enabling `npm run doc` command
Adding callbacks description in doc
Adding snapshot of doc that matching the current status of source code


minggangw (approved):
LGTM, this template really beautifies the html files, thanks!

PR #64, opened 2017-08-14
[ci]Add npm test to the build.sh

qiuzhong:
Add `npm test` to the build.sh so that the test cases will
be executed every time a PR is requested.


minggangw:
As it is always time out, when requesting key B01FA116 from hkp server ha.pool.sks-keyservers.net. So merge this pr first, thanks!

PR #65, opened 2017-08-14
[test]Add test cases to rclnodejs

qiuzhong:
Add 18 test cases from rclpy to rclnodejs.

### DIFF
--- a/test/test-destruction.js
+++ b/test/test-destruction.js
@@ -1,0 +1,110 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const assert = require('assert');
+const rclnodejs = require('../index.js');
+
+describe('Node destroy testing', function() {
+    before(function() {
+        this.timeout(10 * 1000);
+        return rclnodejs.init();
+    });
+
+    after(function() {
+        rclnodejs.shutdown();
+    });
+
+    it('node.destroy()', function() {
+        var node = rclnodejs.createNode('my_node1');
+        node.destory();
+    });
+
+    it('node.destroy() twice', function() {
+        var node = rclnodejs.createNode('my_node1');
+        node.destory();
+
+        assert.doesNotThrow(function() {
+            node.destory();
+        }, Error);
+    });
+
+    it('node.destroy() corrupted', function() {
+        var node = rclnodejs.createNode('my_node3');
+        node.destory();
+
+        var org_handle = node._handle;
+        node._handle = 'garbage'; 
+        try {               
+                node.destory();
+        } catch (TypeError) {
+            assert.ok(true);
+            node._handle = org_handle;
+            node.destory();
+        }
+    });
+
+    it('node.destory timers', function() {
+        var node = rclnodejs.createNode('my_node4');
+        var timer1 = node.createTimer(0.1, () => {}),
+            timer2 = node.createTimer(1, () => {});
+
+        assert.deepStrictEqual(2, node._timers.length);
+        node.destory();
+        assert.deepStrictEqual(0, node._timers.length);
+    });
+
+    it('node destory entities', function() {
+        var node = rclnodejs.createNode('my_node5');
+        
+        var timer = node.createTimer(0.1, () => {});
+        assert.deepStrictEqual(1, node._timers.length);
+        
+        var int16 = rclnodejs.require('std_msgs').msg.Int16;
+        var pub1 = node.createPublisher(int16, 'pub1_topic');
+        assert.deepStrictEqual(1, node._publishers.length);
+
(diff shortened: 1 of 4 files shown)


test/test-destruction.js
```diff
@@ -0,0 +1,110 @@
...
+        return rclnodejs.init();
+    });
+
+    after(function() {
+        rclnodejs.shutdown();
+    });
+
+    it('node.destroy()', function() {
+        var node = rclnodejs.createNode('my_node1');
+        node.destory();
```
Honry: Please use the correct spelling.

test/test-destruction.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    it('node.destroy() corrupted', function() {
+        var node = rclnodejs.createNode('my_node3');
+        node.destory();
+
+        var org_handle = node._handle;
+        node._handle = 'garbage'; 
+        try {               
+                node.destory();
```
Honry: Indent error.

Honry: Please use two whitespace indent as common convention in JavaScript.

test/test-node.js
```diff
@@ -27,8 +115,89 @@ describe('rclnodejs node test suite', function() {
...
+    }, /unknown substitution/, 'Failed to createClient');
+  });
+
+  it('node.createService', function() {
+    var GetParameters = rclnodejs.require('rcl_interfaces').srv.GetParameters;
+    node.createService(GetParameters, 'get/parameters', () => {
+      assert.ok(true);
+    });
+
+    assert.throws(function() {
```
Honry: Move invalid names to an array and use `foreach` to invoke assertion to make test robust.

minggangw:
Please add `mocha` as `devDependencies` in package.json to restrain the ci failure.

test/test-destruction.js
```diff
@@ -0,0 +1,110 @@
...
+// limitations under the License.
+
+'use strict';
+
+const assert = require('assert');
+const rclnodejs = require('../index.js');
+
+describe('Node destroy testing', function() {
+  before(function() {
+    this.timeout(10 * 1000);
```
minggangw: As the performance of the machine used by travis-ci seems not as good as our dev one, the 10s time out is not enough to generate all the necessary js files. I suggest set the time out value to 60s as what I have done on the development branch.

test/test-init-shutdown.js
```diff
@@ -0,0 +1,101 @@
...
+// limitations under the License.
+
+'use strict';
+
+const assert = require('assert');
+const rclnodejs = require('../index.js');
+
+describe('Node destroy testing', function() {
+  before(function() {
+    this.timeout(10 * 1000);
```
minggangw: ditto

test/test-node.js
```diff
@@ -18,6 +18,90 @@ const assert = require('assert');
 const rclnodejs = require('../index.js');
 
 describe('rclnodejs node test suite', function() {
+  describe('createNode method testing', function() {
+    before(function() {
+      this.timeout(10 * 1000);
```
minggangw: ditto

test/test-node.js
```diff
@@ -18,6 +18,90 @@ const assert = require('assert');
...
+    });
+  });
+});
+
+describe('rcl node methods testing', function() {
+  var node;
+  var rclString;
+
   before(function() {
     this.timeout(10 * 1000);
```
minggangw: ditto

test/test-init-shutdown.js
```diff
@@ -0,0 +1,101 @@
...
+      assert.ok(false);
+      done(err);
+    });        
+  });
+
+  it('rclnodejs double init', function(done) {
+    rclnodejs.init().then(function() {
+      assert.ok(true);    
+    }).then(function() {
+      rclnodejs.init();
```
minggangw: As the `init` method returns a Promise, you have to catch the error here OR return it to keep all the promises in the same chain.

test/test-init-shutdown.js
```diff
@@ -0,0 +1,101 @@
...
+      assert.ok(false);
+      done();
+    }).catch(function(err) {
+      assert.ok(true);
+      rclnodejs.shutdown();
+      done(err);
```
minggangw: Pass the `err` will fail the test although the result is expected.

minggangw (commented):
After PR#69 is merged, and the problems here are fixed, all the unit tests can run correctly as I have run them locally. Would you please merge these three patches on this branch together when updating? thanks!

test/test-init-shutdown.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+  });
+
+  it('rclnodejs create node without init', function() {
+    assert.throws(function() {
+      rclnodejs.createNode('my_node');
+    }, /has not been called/, 'Failed to createNode');
+  });
+});
```
minggangw, line 98: invalid ending

test/test-destruction.js
qiuzhong: Sure. I'll update it.

test/test-init-shutdown.js
qiuzhong: Do you mean the `shutdown()` method is not called for cleanup?

test/test-init-shutdown.js
qiuzhong: Yes. I updated but it still got failed. Do you have any good suggestion?

test/test-init-shutdown.js
qiuzhong: Thanks! I'll update it.

test/test-init-shutdown.js
qiuzhong: Got it. I should not pass `err` the `done()`.

minggangw (approved):
LGTM, thanks for the patch which makes the model more robust!

PR #71, opened 2017-08-22
[test]Add test cases for checking members and parameters

qiuzhong:
Add 30 test cases for checking the rclnodejs members and API parameters


minggangw (approved):
LGTM, thanks for the patch!

PR #75, opened 2017-08-23
Make it able to build on Mac

kenny-y:
Make it be able to build, but with execution exception at some cases.
At least timer is able to run on macOS 10.12.6


minggangw (approved):
LGTM, thanks!

PR #78, opened 2017-08-28
[QoS]Add the parameter of QoS for publisher/subscription examples

### LINKED ISSUE #76: QoS-related examples
QoS feature has been enabled in the client/service, publish/subscription APIs. We need some examples for the features.

minggangw:
The publisher/subscription/client/service use the default QoS value as
the parameter when being created.

This patch wants to demonstrate how to use customized parameter of QoS
to create the entities in rclnodejs.

Fix #76

### DIFF
--- a/example/publisher-qos-example.js
+++ b/example/publisher-qos-example.js
@@ -18,13 +18,21 @@
 'use strict';
 
 const rclnodejs = require('../index.js');
+const {QoS} = rclnodejs;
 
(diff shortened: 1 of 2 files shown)


qiuzhong:
Does this update affect the behavior of the publisher/subscription example?

minggangw:
No, it doesn't, just specify the qos parameter, which will be used by the middle ware. The logic is still the same as before.

PR #102, opened 2017-09-18
[test] Fix the Python nodes output redirection issue

qiuzhong:
* This patch fixes the output redirection issue of Python RCL node
launched by `ros2 run demo_nodes_py listener`.
* Add 1 test case


qiuzhong:
I'll recommit this PR.

PR #104, opened 2017-09-18
[test] Add 1 test for Python node output pipe

qiuzhong:
* Fix the Python RCL node output pipe issue
* Add 1 test case


minggangw:
As the circleci hasn't worked on macos platform, so just merge this pr, and thanks for the patch!

PR #105, opened 2017-09-19
Add 28 primitive message type test cases

qiuzhong:
Add 28 test cases for primitive message type checking.


minggangw:
As the circleci hasn't worked on macos platform, so just merge this pr.

PR #109, opened 2017-09-20
[windows]Enable windows build of rclnodejs

### LINKED ISSUE #49: Build on Windows

minggangw:
This patch enables the node-gyp build on windows platform. The tested
platform is:

    Windows8.1 + VS2015 with Update3

Please reference https://github.com/ros2/ros2/wiki/Windows-Development-Setup
to build ROS2 from source, and you have to checkout master branch of it.

Some kind of interactive unit tests will fail at this stage.

Fix #49

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "spdlog"]
 	path = src/third_party/spdlog
 	url = https://github.com/gabime/spdlog.git
+[submodule "src/third_party/dlfcn-win32"]
+	path = src/third_party/dlfcn-win32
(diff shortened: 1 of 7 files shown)


qiuzhong:
Tested on
```
Windows 10 + VS2015 with Update3
```

ros2 can be built successfully and rclnodejs can be built. However, the following issues should be fixed:
* `npm run lint`: this uses `wget` utility, which is not installed by `choco` during in building ros2.
* rclnodejs depends on thirdparty libraries, so before building rclnodejs, we should run `git submodule update --init --recursive`.

PR #115, opened 2017-09-25
Support to pass primitive type directly to publisher/client

### LINKED ISSUE #50: Support a plain JavaScript object as a ROS topic/request
Like `publisher.publish('hello world')`

minggangw:
Currently, when we want to pass a primitive type of message to publisher
or client, we have to create a new class object of that type and assign
the value to its 'data' property, which makes the whole process redundant.

This patch lets user publish primitive message directly, the transition
from primitive message to the corresponding object will occur
implicitly. So you can write code like below:

  const publisher = node.createPublisher(String, 'topic');
  publisher.publish('hello world');

Besides, we also update the example to demonstrate the new usage for
this change.

Fix #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ After building this module, you just need to follow the normal way to use it as 
 
 ```javascript
 const rclnodejs = require('../index.js');
-let String = rclnodejs.require('std_msgs').msg.String;
 
 rclnodejs.init().then(() => {
+  let String = rclnodejs.require('std_msgs').msg.String;
   const node = rclnodejs.createNode('publisher_example_node');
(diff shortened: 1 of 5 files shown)


qiuzhong:
Are all JavaScript primitive types supported or just `String`?

minggangw:
Here the 'primitive' means the message type in ROS2, which is gotten when parsing the .msg files. You can reference https://github.com/RobotWebTools/rclnodejs/blob/develop/rosidl_gen/generator_primitive.js#L40 to checkout the list of primitive types.

qiuzhong:
Oh. I misunderstood that. Thanks for the clarification.

PR #118, opened 2017-09-25
[UT]Fix some issues to pass the tests

minggangw:


qiuzhong:
@minggangw , what's the consideration of this patch? Especially putting the `rclnodejs.shutdown()` to the callback from `after()` to the each test case?

minggangw:
As #120 will fix this issue from root, close this one.

Issue #120, opened 2017-09-27
[UT]Failed to kill ROS2 demo processes


qiuzhong:
With the latest ROS2 code, the ROS2 demo processes launched by commands like `ros2 run demo_nodes_cpp talker` cannot be killed. Any unit test cases that need launch these commands will fail. This happens in [test-cross-lang](https://github.com/RobotWebTools/rclnodejs/blob/develop/test/test-cross-lang.js). #118 doesn't fix this completely.

PR #122, opened 2017-09-27
[test]Fix issue-120

qiuzhong:
Launch the cpp or Python nodes by using the real ros2 cpp demo
executables or python3 to launch rather than using `ros2 run`
command way so the processes can be killed after testing.

Fix:#120


minggangw (approved):
Good job! All the cases passed, thanks! LGTM

PR #126, opened 2017-09-29
[UT]Add a black list of test cases for macOS

qiuzhong:
Add a black list of test cases for macOS so it will not be run
every time a PR is submitted.


minggangw:
@qiuzhong could you also enable the ut on **circleci** if there is no error. You can change the following section of `circle.yml`:
```
test:
  override:
- echo "disable tests"
```

qiuzhong:
@minggangw , can you tell me how to change it?

minggangw:
That's fine, I am going to change it, thanks!

PR #133, opened 2017-10-09
[ut]Avoid calling method done() multi-times

minggangw:
When running test, test-message-type.js, sometimes the call() method
were called several times, which caused the mocha reported an error.

So we'd better call rclnodejs.shutdown() immediately after the call() is
executed to stop reveiving the subscribed message any longer.

### DIFF
--- a/test/test-message-type.js
+++ b/test/test-message-type.js
@@ -35,12 +35,12 @@ const rclnodejs = require('../index.js');
 // }
 
 describe('Rclnodejs message type testing', function() {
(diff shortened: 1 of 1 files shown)


minggangw:
Hi @qiuzhong would you help to confirm this problem, I meet this recently, not every time I run `npm test`, it appears frequently. If this patch does fix it, I will merge the pr then, thanks!

minggangw:
As #135 has made this change, close this one.

PR #137, opened 2017-10-10
[rosidl_gen]Get the property of array type from the wrapper object

### LINKED ISSUE #110: *ArrayWrapper cannot be used to createPublisher
Take the `Byte` as example, there is a `ByteArrayWrapper` class in the `generated/std_msgs/std_msgs__msg__Byte.js`. It stands for an array of the `Byte` message type. However, it cannot be used to `createPublisher`. This code
```
const rclnodejs = require('./rclnodejs/index.js');

rclnodejs.init().then(() => {
  var node = rclnodejs.createNode('byte_publisher');
  var Byte = rclnodejs.require('std_msgs').msg.Byte;
  var ByteArray = Byte.ArrayType;
  var msg = new ByteArray(3);
  msg.setWrappers([1, 2, 3]);
  // console.log(`${msg.data}`);
  console.log(msg.data[0].data); // print 1
  var publisher = node.createPublisher(ByteArray, 'chatter');
  setInterval(() => {
    console.log(`Publishing: ${msg.data}`);
    publisher.publish(msg);
  }, 500);
  rclnodejs.spin(node);
}).catch((err) => {
  console.log(err);
});

```
will lead a `TypeError` when calls `createPublisher`:
```
1
TypeError: typeClass.type is not a function
    at Function.createPublisher (/home/qiuzhong/Workspace/Work/ros2/tmp/rclnodejs/lib/publisher.js:57:26)
    at Node.createPublisher (/home/qiuzhong/Workspace/Work/ros2/tmp/rclnodejs/lib/node.js:111:31)
    at rclnodejs.init.then (/home/qiuzhong/Workspace/Work/ros2/tmp/publisher.js:106:24)
    at process._tickCallback (internal/process/next_tick.js:109:7)
    at Module.runMain (module.js:606:11)
    at run (bootstrap_node.js:389:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:504:3

```

minggangw:
Currently when getting the property of array type, we will get the value
from ref-struct object directly, which has not been assigned.

To fix this, we must get/set the property through the wrapper
consistently.

Fix #110

### DIFF
--- a/rosidl_gen/templates/message.dot
+++ b/rosidl_gen/templates/message.dot
@@ -214,6 +214,14 @@ class {{=objectWrapper}} {
       wrapper._refObject = this._wrapperFields.{{=field.name}}._refArray[index];
       wrapper.deserialize();
     });
+    {{?? field.type.isPrimitiveType && field.type.isArray}}
+    let {{=field.name}}Primitives = this._refObject.{{=field.name}}.data;
+    {{=field.name}}Primitives.length = this._refObject.{{=field.name}}.size;
+    let {{=field.name}}Values = [];
+    {{=field.name}}Primitives.toArray().forEach(value =>{
+      {{=field.name}}Values.push(value.data);
+    });
+    this._wrapperFields.{{=field.name}}.setWrappers({{=field.name}}Values);
     {{?? !field.type.isArray && field.type.type === 'string' && it.spec.msgName !== 'String'}}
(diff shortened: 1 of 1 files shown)


qiuzhong:
Tested and worked! No crash any more!

minggangw:
Thanks so much for your verification, merging...

PR #162, opened 2017-11-02
Run the cpp/py client sequentially

minggangw:
Originally, we launched the cpp and python client simultaneously, but it
seems that the request is never sent on Windows CI(appveyor)
which leads a timeout of  60000ms. So we will launch the two clients
(cpp and python) one by one.


minggangw:
@qiuzhong I meet this problem on appveyor (https://ci.appveyor.com/project/minggangw/rclnodejs/build/develop-133)  recently after e943cfe7 and windows only, I submit this pr to change the way of launching the client, would you please have a look to see whether it's reasonable?

qiuzhong:
@minggangw , I met this problem too when I submitted last PR for the first time CI on Windows. Linux and macOS were fine. However, for the second CI all passed. I have no idea why this happened.

qiuzhong:
@minggangw , it's OK for me so long as the CI can pass easily!

minggangw:
Thanks, I will merge the pr soon.

PR #172, opened 2017-11-10
Remove the execution from build.sh script, and only run istanbul on

minggangw:
Circle on macOS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
   "scripts": {
     "install": "node-gyp rebuild",
     "docs": "cd docs && make",
-    "test": "node ./scripts/compile_cpp.js && istanbul cover ./scripts/run_test.js --report lcovonly",
-    "coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
+    "test": "node ./scripts/compile_cpp.js && node ./scripts/run_test.js",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "install": "node-gyp rebuild",
     "docs": "cd docs && make",
-    "test": "node ./scripts/compile_cpp.js && istanbul cover ./scripts/run_test.js --report lcovonly",
+    "test": "node ./scripts/compile_cpp.js && node ./scripts/run_test.js",
     "coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
```
qiuzhong: This `coverage` npm scritps may not be able to execute on Windows:
* Some Windows users may have no `cat` utility by default.
* `./node_modules/coveralls/bin/coveralls.js` is not an executable, what about `coveralls` simply as `coveralls.cmd` is in `.\node_modules\.bin` and this path is loaded by npm scripts automatically.
* `rm`, same as `cat`

minggangw: As we only need the `coverage` on ci to upload the result to overalls.io, we can run it on a specific ci vendor, e.g. deploy on .travis, to avoid executing on different platforms. what do you think @qiuzhong ?

minggangw: I mean we can remove `npm run coverage` from package.json, and run it through the ci script.

qiuzhong: It's OK for me to remove it.

PR #180, opened 2017-11-17
Fix the memory leak cases

minggangw:
This patch mainly fixed two memory leak circumstances:

1. When calling rcl_take_response, a pointer of type of rmw_request_id_t
is passed, but never gets deallocated. We should free the memory
immediately.

2. After allocating a pointer of rmw_qos_profile_t, we have to
deallocate it. So a unique_ptr will take the ownship of it.

These problems above should not be the root cause of issue#175.

### DIFF
--- a/src/rcl_bindings.cpp
+++ b/src/rcl_bindings.cpp
@@ -25,6 +25,8 @@
 #include <rmw/validate_namespace.h>
 #include <rmw/validate_node_name.h>
 #include <rosidl_generator_c/string_functions.h>
+
+#include <memory>
 #include <string>
 
 #include "handle_manager.hpp"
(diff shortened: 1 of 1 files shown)


minggangw:
There is a build error when compiling `rclcpp`, an [issue](https://github.com/ros2/rclcpp/issues/398) has been submitted.

minggangw:
Hi @qiuzhong would you please take a look at the ut on windows ci platform, the errors are below:

```
Uncaught Error: spawn C:\proj\rclnodejs\test\cpp\publisher_msg ENOENT
```

and

```
Uncaught Error: spawn C:\proj\rclnodejs\test\cpp\subscription_msg ENOENT
```

qiuzhong:
Looks like the cpp executables didn't exists. Maybe failed to build. I'm investigating.

qiuzhong:
The test cpp code depends on `rclcpp`, which was not built successfully in the CI due to  [issue](https://github.com/ros2/rclcpp/issues/398). This is true for macOS and Linux.

minggangw:
Thanks, let's wait the upstream to update the code, it should be fixed soon. Then re-run the ci again.

qiuzhong:
However, the Windows CI uses a built zip package and it should be another case, I downloaded this package and found it doesn't work at all at `ament build . ` phase.

PR #182, opened 2017-11-20
Add python dependencies to fix the windows ci error

minggangw:


appveyor.yml
```diff
@@ -31,7 +31,10 @@ before_build:
   - setx AMENT_PYTHON_EXECUTABLE "c:\Python36-x64"
```
qiuzhong: PYTHON3 path has been changed. So I guess it should be updated here as well.

PR #195, opened 2017-11-23
Support to require a single type of message/service

### LINKED ISSUE #179: Consider a new form of rclnodejs.require() to save disk reading cost
Currently we have to use

```
const all = rclnodejs.require('geometry_msgs'); // Lots of high-cost sync disk reading
const whatIWant = all.msgs.Twist; // The only thing that user needs
```

```
const whatIWant = rclnodejs.require('geometry_msgs', 'msg', 'Twist');
 - OR -
const whatIWant = rclnodejs.require('geometry_msgs/msg/Twist');
```

minggangw:
Currently if we want to require a type of message, you have to get the
whole messages of that package, which will cost a lot of time because of
the sync manipulation.

e.g. If we want to require the String class, and write:
    // This gets all stuff included in the std_msgs package.
    const String = rclnodejs.require('std_msgs').msg.String;

To improve the performance, this patch enables to require a single type
of message which belongs to a specific package.

e.g. You can get the String class as follows:
    const String = rclnodejs.require('std_msgs/msg/String');

Fix #179


qiuzhong:
@minggangw , do we have to adjust the usage for all message types?

minggangw:
Not necessary, the new `require` method is compatible with the previous implement. I suggest adding a new test case to verify this patch, something like:

```JavaScript
const String = rclnodejs.require('std_msgs/msg/String');
```

to see if it works fine, and that's enough.

qiuzhong:
OK, I'll add that case.

PR #197, opened 2017-11-23
Test cases for extra destroy methods

qiuzhong:
Test cases for publisher/subscription/client/service/timer destroy methods.


minggangw:
LGTM, merge.

PR #200, opened 2017-11-24
Test cases for new style require messages or services

qiuzhong:
Add 2 test cases for new style require messages, like this:
const String = rclnodejs.require('std_msgs/msg/String')


minggangw:
We raised another 1% in coverage 👍 thanks! LGTM

PR #210, opened 2017-11-30
Change of API (Issue #189)

### LINKED ISSUE #189: Change of API: hide ROSMessageWrapper type
Purpose:
 - Remove `rclnodejs.require()` from customer-visible API
 - Use string directly to create publisher/subscriber/client/service
 - Provide a function that can create a blank message/request JavaScript object for later use
 - Add Response class for writing response to client

Example code 1:

```
  let obj = rclnodejs.createMessageObject('sensor_msgs/msg/JointState');
  obj.name = ['Alice', 'Bob'];
  obj.position = [0, 1];
  // obj.xxx = ...

  const publisher = node.createPublisher({package: 'sensor_msgs',
                                          type: 'msg',
                                          name: 'JointState'
                                          }, 'topic_name');
   -- OR --
  const publisher = node.createPublisher('sensor_msgs/msg/JointState', 'topic_name');

  publisher.publish({name: ['Alice', 'Bob'], position: [0, 1], /*...*/});
   -- OR --
  publisher.publish(obj);
```

Example code 2:

```
const msgType = {
  package: 'std_msgs',
  type: 'msg',
  name: 'String',
};

node.createService(msgType, ...);
node.createClient(msgType, ...);
node.createPublisher(msgType, ...);
node.createSubscription(msgType, ...);
```

Example code 3:
```
const rclnodejs = require('rclnodejs');

rclnodejs.init().then(() => {
  const node = rclnodejs.createNode('publisher_message_example_node');
  const publisher = node.createPublisher('sensor_msgs/msg/JointState', 'TopicJointState');
  let count = 0;

  setInterval(function() {
    publisher.publish({
      header: {
        stamp: {
          sec: 123456,
          nanosec: 789,
        },
        frame_id: 'main frame',
      },
      name: ['Tom', 'Jerry'],
      position: [1, 2],
      velocity: [2, 3],
      effort: [4, 5, 6],
    });
    console.log(`Publish ${++count} messages.`);
  }, 1000);

  rclnodejs.spin(node);
});
```

Example code 4:

```
const rclnodejs = require('rclnodejs');

rclnodejs.init().then(() => {
  let node = rclnodejs.createNode('service_example_node');

  node.createService('example_interfaces/srv/AddTwoInts', 'add_two_ints', (request, response) => {
    console.log(`Incoming request: ${typeof request}`, request);
    let result = response.template;
    result.sum = request.a + request.b;
    console.log(`Sending response: ${typeof result}`, result, '\n--');
    response.send(result);
  });

  rclnodejs.spin(node);
});
```

kenny-y:
Fixes #189

Hide ROS message wrapper class in API, use simple string or plain JS object as parameter.

Also adding Response class to request callback of service.


kenny-y:
```
  445 passing (2m)
  1 failing

  1) Cross-language interaction Node.js service Node.js service should work with C++ client:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
This is common on slow CI server... will restart travis again manually.

test/test-single-process.js
```diff
@@ -72,23 +71,78 @@ describe('Test rclnodejs nodes in a single process', function() {
...
+      response.send({sum: request.a + request.b});
+    });
+    var client = clientNode.createClient(AddTwoInts, 'single_ps_channel2_3');
+    const request = {a: 1, b: 2};
 
     var timer = clientNode.createTimer(100, () => {
       client.sendRequest(request, (response) => {
         timer.cancel();
         assert.deepStrictEqual(response.sum, 3);
+        serviceNode.destroy();
```
qiuzhong, line 144: Thanks for the destruction code!

lib/client.js
```diff
@@ -17,6 +17,7 @@
 const rclnodejs = require('bindings')('rclnodejs');
 const Entity = require('./entity.js');
 const debug = require('debug')('rclnodejs:client');
+const {MessageTranslator} = require('./message_translator.js');
```
minggangw: As Publisher/Subscription/Client/Service will use the `MessageTranslator`  collectively, I suggest moving it to their base class `Entity`, and use a variable name beginning with `_` (`this._translator`) to indicate we will use it privately.

lib/message_translator.js
```diff
@@ -102,8 +161,31 @@ class MessageTranslator {
...
+  toROSResponse(obj) {
+    let msg = new this._typeClass.Response();
+    const type = typeof obj;
+    if (type === 'string' || type === 'number' || type === 'boolean') {
+      msg.data = obj;
+    } else if (type === 'object') {
+      copyMsgObject(msg, obj);
+    }
+    return msg;
+  }
```
minggangw: I suggest combining `toROSRequest`, `toROSResponse` and `toROSMessage` together, it's better not to have translator to own a `typeClass` of a specific message, instead we can pass the `typeClass` as a parameter:

```
static toROSMessage(typeClass, obj) {...
```

lib/interface_loader.js
```diff
@@ -44,6 +74,16 @@ let interfaceLoader = {
   },
 
   loadInterface(packageName, type, messageName) {
```
minggangw, line 76: We have the same logic in https://github.com/RobotWebTools/rclnodejs/blob/develop/index.js#L170, and we are going to hide the `require` method, I suggest using this implementation to replace the one of `require` in `index.js`.

lib/service.js
```diff
@@ -17,6 +17,44 @@
...
+        Note: you can use .template to get an empty result object.
+   * @return {undefined}
+   * @see {@link Response#template}
+   */
+  send(response) {
+    const rawROSResponse = this._service.translator.toROSResponse(response).serialize();
+    rclnodejs.sendResponse(this._service._handle, rawROSResponse, this._header);
+    this.sent = true;
+  }
+}
```
minggangw: I didn't see the corresponding class, something like `class Request`, in `lib/client.js`, is it necessary here?

minggangw (commented):
Other code is good to me, please @qiuzhong have a look at the examples and tests, thanks!

lib/interface_loader.js
kenny-y: Will log an issue later to track this. Currently `require()` can do more than `loadInterface()` and there are test cases relying on it.

lib/client.js
kenny-y: Will remove member `this._translator`, see below.

lib/message_translator.js
kenny-y: Will drop `MessageTranslator` class and use 2 direct functions.

kenny-y:
Ping @minggangw @qiuzhong :)

minggangw:
We can merge this PR now to make a step to the next release :+1:

qiuzhong:
The code change for examples and unit tests is good for me. In fact, I found some little issues (Like inconsistent variable naming or missing the node destruction) in the unit tests. Anyway, I will submit next PRs to fix them.

kenny-y:
I also noticed mixed usage of `var` `let` `const`, better way is to
 - Discard `var` and use `let` and `const`
 - Use `const` as much as possible.

qiuzhong:
@kenny-y , thanks. However I meant that the some other issues that already existed in unit tests before your update.

PR #213, opened 2017-11-30
Consistent message String type naming and timer for publisher

qiuzhong:
* Use RclString naming for all std_msgs/msg/String type
* For all publishers/clients, use the `node.createTimer` timer from rcl
* Add node destruction before rclnodejs shutdown


minggangw:
Merging...

kenny-y:
Impressive quick move on fixing small defects :+1: :+1: :+1:

PR #219, opened 2017-12-01
Use TypedArray for primitive array so that it can save memory and run faster at the same time

kenny-y:
There is still possibility to deeper engage continuous memory in primitive ROS array (like what `rclcpp` does), but that's another story and will be tracked in another issue.


minggangw:
Thanks for the patch, merging

PR #229, opened 2017-12-11
Enable the cross language tests

minggangw:
As 2175273a7c4a778991 has landed, it is the time to enable these tests
to make the program robust.


minggangw:
@qiuzhong would please take a look at the errors which happened on windows ci?

qiuzhong:
The errors look like before. I'm afraid I can't get a clue from the CI error log.

minggangw:
As we will refactor this kind of test, close this one.

Issue #235, opened 2017-12-12
Error Reported When Providing An Empty Object Array


kenny-y:
Test Case
```
// sensor_msgs/msg/PointCloud2:
  const data = {
    header: {
      stamp: {sec: 11223, nanosec: 44556},
      frame_id: 'f001',
    },
    height: 240,
    width: 320,
    fields: [], // This is the empty array
    is_bigendian: false,
    point_step: 16,
    row_step: 320*16,
    data: uint8Data,
    is_dense: false,
  },
```
The error:
```
TypeError: writeUInt64: Number/String 64-bit value required
      at TypeError (native)
      at Buffer.writeUInt64LE (node_modules/ref/lib/ref.js:1376:18)
      at Object.set (node_modules/ref/lib/ref.js:973:53)
      at Object.set (node_modules/ref/lib/ref.js:487:10)
      at StructType.desc.set [as size] (node_modules/ref-struct/lib/struct.js:225:16)
      at PointFieldArrayWrapper._resize (generated/sensor_msgs/sensor_msgs__msg__PointField.js:188:26)
      at PointFieldArrayWrapper.copy (generated/sensor_msgs/sensor_msgs__msg__PointField.js:209:10)
      at PointCloud2Wrapper.fields (generated/sensor_msgs/sensor_msgs__msg__PointCloud2.js:147:32)
      at copyMsgObject (lib/message_translator.js:48:20)
      at Object.toROSMessage (lib/message_translator.js:159:5)
      at Context.<anonymous> (test/test-array.js:308:32)
```

Providing `fields: [{}]` can work around this issue.

Issue #234, opened 2017-12-12
Performance: make sure large arrays are all processed effeciently


kenny-y:
Array usage samples:

```
Searching 162 files for "[]"

ros2_ws/ros2_source/src/ros2/common_interfaces/actionlib_msgs/msg/GoalStatusArray.msg:
    2  # by an action server
    3  std_msgs/Header header
    4: GoalStatus[] status_list
    5  

ros2_ws/ros2_source/src/ros2/common_interfaces/diagnostic_msgs/msg/DiagnosticArray.msg:
    1  # This message is used to send diagnostic information about the state of the robot.
    2  std_msgs/Header header # for timestamp
    3: DiagnosticStatus[] status # an array of components being reported on
    4  

ros2_ws/ros2_source/src/ros2/common_interfaces/diagnostic_msgs/msg/DiagnosticStatus.msg:
   16  string hardware_id
   17  # An array of values associated with the status.
   18: KeyValue[] values
   19  
   20  

ros2_ws/ros2_source/src/ros2/common_interfaces/diagnostic_msgs/srv/SelfTest.srv:
    2  string id
    3  byte passed
    4: DiagnosticStatus[] status
    5  

ros2_ws/ros2_source/src/ros2/common_interfaces/geometry_msgs/msg/Polygon.msg:
    1  # A specification of a polygon where the first and last points are assumed to be connected
    2  
    3: Point32[] points
    4  

ros2_ws/ros2_source/src/ros2/common_interfaces/geometry_msgs/msg/PoseArray.msg:
    3  std_msgs/Header header
    4  
    5: Pose[] poses
    6  

ros2_ws/ros2_source/src/ros2/common_interfaces/nav_msgs/msg/GridCells.msg:
    4  float32 cell_width
    5  float32 cell_height
    6: geometry_msgs/Point[] cells
    7  

ros2_ws/ros2_source/src/ros2/common_interfaces/nav_msgs/msg/OccupancyGrid.msg:
    8  # The map data, in row-major order, starting with (0,0).  Occupancy
    9  # probabilities are in the range [0,100].  Unknown is -1.
   10: int8[] data
   11  

ros2_ws/ros2_source/src/ros2/common_interfaces/nav_msgs/msg/Path.msg:
    2  
    3  std_msgs/Header header
    4: geometry_msgs/PoseStamped[] poses
    5  

ros2_ws/ros2_source/src/ros2/common_interfaces/sensor_msgs/msg/BatteryState.msg:
[...]

Issue #233, opened 2017-12-12
Generated messages: some methods should be static


kenny-y:
``` JavaScript

  isPrimitive() {
    return true;
  }

  get isROSArray() {
    return true;
  }

  // Use this to get class type to access all static functions/properties
  get classType() {
    return Int64Wrapper; // Taking Int64 for example
  }

```

kenny-y:
Mark this as low priority manually...

Issue #232, opened 2017-12-12
Failed to Assign a New Value to an Array Element


kenny-y:
The following test case would fail

``` JavaScript
    const r = Math.round(Math.random() * 100);
    msg1.data[0] = r;
    assert(msg1.data[0] == r);
```

The root cause is that the getter of data property is wrong. We should cache an array in `this` object, clear that cache array when it's necessary, and generate the cache array when requested in `.data` getter when it's empty.

e.g. This the generated code for Int64MultiArray
``` JavaScript
  get data() {
    let values = [];
    this._wrapperFields['data'].data.forEach((wrapper, index) => {
      values.push(wrapper.data);
    });
    return values;
  }
```

Resolution
``` JavaScript
  get data() {
    if (!this.dataArray) {
      this.dataArray = [];
      this._wrapperFields['data'].data.forEach((wrapper, index) => {
        this.dataArray.push(wrapper.data);
      });
    }
    return this.dataArray; // Pass by reference
  }

// TODO: assign undefined to this.dataArray when necessary.
```

TypedArray implementation doesn't have this issue so far.

kenny-y:
@minggangw Would you please look into this? I suppose we should clear `this.dataArray` when doing `.copyRefObject()`, and we should copy `this.dataArray` to `this._wrapperFields['data']` when necessary... is there anything else?

minggangw:
I will take this.

PR #236, opened 2017-12-12
Use TypedArray to achieve good performance

### LINKED ISSUE #191: Make sure large data are represented in TypedArray, and also make sure there is no deep copy
Purpose:
 - Make sure the data is represented in `TypedArray` (instead of JavaScript `Array`)
 - No deep copy when copying messages

e.g. in `sensor_msgs/msg/Image.msg`

```
uint8[] data          # actual matrix data, size is (step * rows)
```

### LINKED ISSUE #220: Make It Efficient: Primitive Type ROS Array
We should use direct `type* buffer` for primitive type ROS array bases on the below findings.

image

kenny-y:
Fix #191 
Fix #220

### DIFF
--- a/src/rcl_bindings.cpp
+++ b/src/rcl_bindings.cpp
@@ -696,6 +696,10 @@ NAN_METHOD(InitString) {
     info.GetReturnValue().Set(Nan::Undefined());
 }
 
+inline char* GetBufAddr(v8::Local<v8::Value> buf) {
+  return node::Buffer::Data(buf.As<v8::Object>());
+}
+
 NAN_METHOD(FreeMemeoryAtOffset) {
   v8::Local<v8::Value> buf = info[0];
   if (!node::Buffer::HasInstance(buf)) {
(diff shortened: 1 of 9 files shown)


kenny-y:
Known issues were logged as #232 #233 #234 #235

src/rcl_bindings.cpp
```diff
@@ -696,6 +696,10 @@ NAN_METHOD(InitString) {
     info.GetReturnValue().Set(Nan::Undefined());
 }
 
+inline char* get_addr(v8::Local<v8::Value> buf) {
```
minggangw: As `get_addr` doesn't return a attribute of a class, like set/get pair, shall we use `GetAddress` to make the convention consistent. Others are good to me.

kenny-y:
@qiuzhong what do you think of the changes of the tests? :)

minggangw (approved):

qiuzhong:
@kenny-y , the changes are OK for me. But I would suggest you to put your added code of `test-array.js` in another test file because they're testing differently aspects. My code mainly focus on array data communication and your code is focus on array data verification.

PR #238, opened 2017-12-14
Return a plain JavaScript object in rclnodejs.createMessageObject().

### LINKED ISSUE #237: Return value of rclnodejs.createMessageObject() is not translated to plain JavaScript object
Also the arrays members are empty, should we fill in a placeholder element, esp. when it's an object?

kenny-y:
This is used to get a view of the message structure, name of fields, types and etc.

Fix #237


minggangw (commented):
LGTM, merging.

PR #246, opened 2017-12-21
Handle the SIGINT

### LINKED ISSUE #244: No implicit signal handler for rclnodejs
Well, for `rclcpp` and `rclpy`, if a process is in the event loop, pressing `Ctrl + C` will raise `signal_handler(2)` and `KeyboardInterrupt` and then get the return value `0` and `1` respectively. So this means the `rclcpp` and `rclpy` already handle `SIGINT` by default, users don't need to write explicit handler for cleanup. 

But for a `rclnodejs` process, pressing `Ctrl + C` will simply terminate the process and return `130` code. It seems there is no such cleanup(like `node.destroy()` or `rclnodejs.shutdown()`)

minggangw:
When rclnodejs receives the SIGINT, it must be handled correctly to
release the resources and quit.

Fix #244

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,4 +239,10 @@ let rcl = {
   },
 };
 
+process.on('SIGINT', () => {
+  debug('Catch ctrl+c event and will cleanup and terminate.');
+  rcl.shutdown();
+  process.exit(0);
+});
+
 module.exports = rcl;


qiuzhong:
Tested and worked. But the log is displaying in debug mode only. In release mode, without any prompt message before termination.

Issue #222, opened 2017-12-05
Cross-language interaction failed


minggangw:
Recently when running script:
```
node ./scripts/compile_cpp.js && node ./scripts/run_test.js
```

The appveyor ci fails consistently, please see pr #221, and I noticed that the failed cases are the ones which test the compatibility between `rclnodejs` and `rclpy`/`rclcpp` clients. We verify the messages by reading what has been written to the std out, which may cause some mistake when the console outputs more. e.g. If we enable the debug build, there will be a lot of debug messages which will also influence the test result.

@qiuzhong would you please change the strategy to verify the interaction related functions. Maybe we could create another `publisher/subscription` in `rclpy/rclcpp` to forward what has been `received/published` back to `rclnodejs` instead of reading from console output.

minggangw:
Using our development machine on Linux, all unit tests can pass. But the performance of the machine used by the ci may be not as high as ours, and we have to eliminate the cases which depend on the timing sequence.

qiuzhong:
@minggangw , indeed, by capturing the recipient standard output is not a perfect way. I wonder in 
`rclnodejs publisher -> rclpy/rclcpp subscription` case, how does the subscription send back the message to the rclnodejs publisher nodes? I don't quite understand how to add more `publisher/subscription` nodes in `rclpy/rclcpp` to forward messages?

minggangw:
We can adopt the policy below:

- In `rclnodejs`, create a publisher; In `rclpy`, create a subscription.
- The publisher of `rclnodejs` sends a message to subscription.
- The subscription receives the message from the `rclnodejs` side, then uses another publisher to send back what it just received.
- Again, in `rclnodejs`, create a subscription to receive message from `rclpy` to verify if the message is equal to what it just sent. 

This can improve the ut pass rate on ci, and not a functional requirement.

qiuzhong:
I tried the new strategy to verify the interaction between publisher/subscription and it works. However, there is another issue: the `rclpy` node is quite slow. If run the `rclnodejs` test cases as soon as I launch the `rclpy` node, the `rclpy` node often fails to response as it may not be initialized. So maybe I should still use the timer to send message from `rclnodejs` publisher to `rclpy` subscription and then send it back from `rclpy` publisher to `rclnodejs` subscription and compare the sent and received message.

qiuzhong:
The following unit test files are using the console output and need to use the new strategy for verification:
* test-cross-lang.js
[...]

PR #250, opened 2017-12-26
Use new strategy to verify message transmitting

qiuzhong:
Before we used console output of C++ and Python subscription to
check the message, this may sometimes fail on low performance machine.
Now we use a new strategy to verify message:
* Rclnodejs publisher sends message to C++ or Python subscriptions
* The subscriptions send the message back by creating a new publisher
* The rclnodejs side then creates a new subscription to receive messages

Therefore, there is no need to keep the rclnodejs subscription - C++/Python
publisher cases any more.

Fix: #222

### DIFF
--- a/scripts/compile_cpp.js
+++ b/scripts/compile_cpp.js
@@ -19,22 +19,40 @@ const os = require('os');
 const path = require('path');
 const child = require('child_process');
 
-var platform = os.platform();
+// var platform = os.platform();
 var rootDir = path.dirname(__dirname);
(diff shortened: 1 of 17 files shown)


qiuzhong:
The failure is due to missing of `add_two_ints_client` executable in CI, which was generated in my local environment. I'm investigating and fixing it.

minggangw:
It's OK, I will merge it when all ci pass, thanks!

minggangw (approved):
Thanks a lot for the great work to improve the test efficiency! We can run full tests now. LGTM

PR #254, opened 2017-12-28
Complete test cases for missing APIs in PR251

qiuzhong:


minggangw:
Thanks for the quick response and merge.

PR #261, opened 2018-01-24
Add nan as a depenency in package.json

minggangw:
From the issue https://github.com/RobotWebTools/ros2-web-bridge/issues/33,
when using node v4.4.2, the nan module is not installed. This bug can
not be reproduced on lts carbon.

This patch adds the nan module explicitly to fix this problem.


qiuzhong:
Tried on Windows and Linux with Node.js `4.4.2`, the installation passed. However, there are ES6 syntax used in `rclnodejs` while not supported by Node.js 4.x and failed to `require('./index.js')`:

```
> require('./index.js');
C:\dev\ros2\bugs\rclnodejs\index.js:82
  createNode(nodeName, namespace = '') {
                                 ^
SyntaxError: Unexpected token =
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at repl:1:1
    at REPLServer.defaultEval (repl.js:262:27)
    at bound (domain.js:287:14)
```

qiuzhong:
@minggangw , what do you think?

minggangw:
Sorry, I didn't run the testing, as we use the `nan` directly, it's unharmful to add it to `package.json` here.

qiuzhong:
I guess we have to drop supports for Node.js 4.x?

minggangw:
Absolutely, would you please find out from which edition of node.js can pass our tests and submit a pr to describe this key requirement in `README`, thanks!

qiuzhong:
OK. I'll update the README.

PR #262, opened 2018-01-24
Add lowest Node.js version requirement

qiuzhong:
The lowest Node.js that can pass the unit tests is 6.10.0.
Add this requirement explicitly in README to avoid kinds of errors.


minggangw (approved):
LGTM, thanks to check this quickly!

Issue #272, opened 2018-02-24
Misusage of assert.ok() in UT


minggangw:
I found a problem in https://github.com/RobotWebTools/rclnodejs/blob/develop/test/test-compound-msg-type-check.js#L62

```JavaScript
    assert.ok(typeof header.stamp, 'string');
    assert.ok(typeof header.frame_id, 'string');
```

I think we should use `assert.deepStrictEqual()` here, @qiuzhong please double check it, thanks!

qiuzhong:
@minggangw , thanks! You're right, I'll correct it soon.

PR #273, opened 2018-02-24
[UT] Fix incorrect assert statements

qiuzhong:
Correct the assert statements in issue 272

Fix: #272


minggangw:
LGTM and thanks!

PR #245, opened 2017-12-19
Add security-related tests for rclnodejs

qiuzhong:
* Destroy non-existent nodes (+6)
* Fuzzing API calls (+9)

Issue #241, opened 2017-12-15
No consistent type checking for string before publish


qiuzhong:
Suppose we create a publish with `std_msg/msg/String` message.
```
var node = rclnodejs.createNode('any_node_name');
var publisher = node.createPublisher('std_msgs/msg/String', 'chatter');

// This doesn't throw a TypeError!
publisher.publish({a: 1});
```
Seems publisher with `string` type can work with `object` type.
But if I publish another inconsistent primitive type like `publisher.publish(255);` then a `TypeError` can be  thrown:
```
TypeError: writeUInt64: Number/String 64-bit value required
```

qiuzhong:
@kenny-y , @minggangw , I was doing some fuzzing API testing with random parameters. And I was using the inconsistent message type on purpose.

qiuzhong:
Not limited to `std_msgs/msg/String`, seems any standard message types can accept `object` as the publishing data.

kenny-y:
Yes, `.publish()` should throw an error when the argument is ill-formed.

I remember this was planned but delayed... I had talked to @minggangw on this topic when I was doing plain message translation (#189), it was tracked in #190

qiuzhong:
Well, `publisher.publish()` can accept JavaScript primitive types or object. The issue is not whether it can accept object while the message is known, but any object data no matter if it's legal or not can be published.

kenny-y:
From my understanding we want to do this:

For `'std_msgs/std/String'`, we want to accept `{data: 'xxx'} OR 'xxx'` as an argument but reject `{what: 129}`
To generalize that we can conclude that the unexpected object structure/value-type should be rejected.

Anything missing here?

qiuzhong:
> For 'std_msgs/std/String', we want to accept {data: 'xxx'} OR 'xxx' as an argument but reject {what: 129}

I agree with you.

And the rule should be applied to all standard message types.

minggangw:
dup with #243

qiuzhong:
Fixed by #274.

PR #274, opened 2018-02-24
Check the consistency of the message's members when being serialized

### LINKED ISSUE #243: Undefined properties of requests are converted to 0 in the callback of createService
Let's say the client sends an ill-formatted request `{a:1}` for the service `example_interfaces/srv/AddTwoInts` (which requires two properties in the request), in the callback of 
`createService`:
```
  node.createService('example_interfaces/srv/AddTwoInts', 'add_two_ints', (request, response) => {
    console.log(`Incoming request: ${typeof request}`, request);
    let result = response.template;
    result.sum = request.a + request.b;
    console.log(`Sending response: ${typeof result}`, result, '\n--');
    response.send(result);
  });
```
We get this output:
```
Incoming request: object { a: 69, b: 0 }
Sending response: object { sum: 69 } 
--

```

Undefined property `request.b` is converted to 0 and doesn't throw an Error. If you print the type of request.b, it's `number`.

minggangw:
Currently when a message is instantiated, the values of its data members will
be a "default". Reference the issue 243:

If we have a request which will be sent to the service to calculate the
sum of two integers, and create {a: 1} instead of {a: 1, b: 1} which
means we do not assign the value for member b explicitly. What the
service will receive is {a: 1, b: 0} from the client request. We may
just forget the assignment unintentionally, but rclnodejs should produce
an error.

This patch implements the strategy below:
1. If the type of a data member in the message is primitive, including
Bool, Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64, UInt64, Float64,
Float32, Char, Byte and String, the data member will be not be created and
undefined is returned when being queried.

2. Other types of the data members, including compound data member and
arrays (primitive & non primitive), it will be created.

If the message is being serialized, which occurs during publishing a topic or
sending a request, the consistency of the message will be checked. Once
one of the data members is undefined, a TypeError will be thrown to
indicate the name of the invalid one.

Fix #243

### DIFF
--- a/lib/message_translator.js
+++ b/lib/message_translator.js
@@ -25,7 +25,7 @@ function isTypedArray(value) {
 function copyMsgObject(msg, obj) {
   if (typeof obj === 'object') {
     for (let i in obj) {
-      if (typeof msg[i] !== 'undefined') {
+      if (msg.hasMember(i)) {
         const type = typeof obj[i];
         if (type === 'string' || type === 'number' || type === 'boolean') {
           // A primitive-type value
--- a/rosidl_gen/generator.json
+++ b/rosidl_gen/generator.json
@@ -1,6 +1,6 @@
 {
   "name": "rosidl-generator",
-  "version": "0.0.7",
+  "version": "0.0.8",
   "description": "Generate JavaScript object from ROS IDL(.msg) files",
   "authors": [
     "Minggang Wang <minggang.wang@intel.com>",
--- a/rosidl_gen/templates/message.dot
+++ b/rosidl_gen/templates/message.dot
@@ -178,6 +178,14 @@ function getPackageNameByType(type) {
     return type.pkgName;
   }
 }
+
+function extractMemberNames(fields) {
+  let memberNames = [];
+  fields.forEach(field => {
+    memberNames.push(field.name);
+  });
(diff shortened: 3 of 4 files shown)


minggangw:
@qiuzhong would you please cherry pick this PR and then apply PR #245 to verify, I think you have to change a little bit of it and add more test cases to check the functionality.

Please **READ** the commit log above which will help you to understand the new implementation and design your tests. Once #245 passes, I will merge these two PRs orderly. Thanks!

qiuzhong:
@minggangw , sure, I'm testing this PR.

qiuzhong:
Verified. #241 and #243 fixed.

minggangw:
Thanks! I will rebase this PR when #245 is updated, and verify it again.

qiuzhong:
@minggangw , I just updated my cases in #245. But it will only pass based on your PR. So please you merge it first?

minggangw:
Can all the updated cases pass locally (with this PR applied)?

qiuzhong:
Yes. It worked for me in the local repository.

minggangw:
Okay, I will merge this one first.

qiuzhong:
Thanks!

PR #282, opened 2018-03-07
Upgrade Node.js to latest LTS Carbon v8.10.0 on CIs

minggangw:


qiuzhong:
Track the failure of Linux in issue #284

PR #283, opened 2018-03-07
Build rclnodejs addon on the binary package of ROS2 

minggangw:


qiuzhong:
@minggangw , the error occurred when running cpplint. Please take a look.

minggangw:
I noticed that already, the circleci deployed python2 & 3 simultaneously which caused this problem and I'm working on it.

minggangw:
Error log

>/usr/local/Cellar/python/3.6.4_3/Frameworks/Python.framework/Versions/3.6/Resources/Python.app/Contents/MacOS/Python: can't open file 'cpplint.py': [Errno 2] No such file or directory

PR #292, opened 2018-03-14
[PnP]Add PnP tests for rclnodejs client (JavaScript)

### LINKED ISSUE #277: Add rclnodejs benchmark test used for PnP

minggangw:
This patch added the tests used for rclnodejs. We defined two kinds of
PnP tests, which are stress testing and endurance testing.

This patch added the tests used for rclnodejs. We defined two kinds of
PnP tests, which are stress testing and endurance testing.

1. Stress testing
The test case is designed to find the behaviour beyond the normal
expected overload, and we can find the breaking points and short term
memory leaks through the tests. There is one test case for each of the
publisher/subscription and client/service. The strategy is:

- A publisher publishes a topic of std_msgs/UInt8MultiArray type every
  100ms, which contains an array of size 10MB, and sends 36000 times
  totally.
- A client sends a request of nav_msgs/GetMap whose response will
  contains an array of size 10MB until it receives the response from the
  service 36000 times.

2. Endurance testing
The test case is designed to find the expected overload for a longer
term, and we can find the long term memory leaks through the tests.
There is one test for each of the publisher/subscription and
client/service. The stragegy is:

- A publisher publishes a topic of std_msgs/JointState type every
  100ms for 24 hours (864000 totally).
- A client sends a request of std_srvs/SetBool until it receives the
  response from the service 864000 times.

All tests begin with a line of output 'Begin' and end with 'End'.

Fix #277

### DIFF
--- a/benchmark/rclnodejs/client-endurance-test.js
+++ b/benchmark/rclnodejs/client-endurance-test.js
@@ -1,0 +1,44 @@
+// Copyright (c) 2018 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('../../index.js');
+
+rclnodejs.init().then(() => {
+  console.log('The client will send a SetBool request continuously until receiving response 864000 times.');
+  console.log(`Begin at ${new Date().toString()}.`);
+
+  const node = rclnodejs.createNode('endurance_client_rclnodejs');
+  const client = node.createClient('std_srvs/srv/SetBool', 'set_flag');
+  let sentTimes = 0;
+  let receivedTimes = 0;
+  let totalTimes = 864000;
+
+  let sendRequest = function(response) {
+    client.sendRequest(true, (response) => {
+      if (++receivedTimes > totalTimes) {
+        rclnodejs.shutdown();
+        console.log(`End at ${new Date().toString()}`);
+      } else {
+        setImmediate(sendRequest);
+      }
+    });
+  };
+
+  sendRequest();
+  rclnodejs.spin(node);
+}).catch((e) => {
+  console.log(`Error: ${e}`);
+});
--- a/benchmark/rclnodejs/client-stress-test.js
+++ b/benchmark/rclnodejs/client-stress-test.js
@@ -1,0 +1,46 @@
+// Copyright (c) 2018 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+/* eslint-disable camelcase */
+const rclnodejs = require('../../index.js');
+
+rclnodejs.init().then(() => {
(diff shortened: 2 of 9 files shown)


qiuzhong:
@minggangw , there are lint errors:

```
benchmark/rclnodejs/publisher-stress-test.js
  33:1  error  Line 33 exceeds the maximum line length of 120  max-len
```

Please fix it first.

minggangw:
I have noticed this error and will correct it, thanks!

minggangw:
Hi @Honry @qiuzhong please take a loot at this pr and combine with #294 and #298 together, which are the PnP tests for `rclnodejs` to compare the Node.js with Python/C++. Please read the commit log about the design principle and consider the factors below while reviewing:

- Rationality - whether these cases can achieve the goal that the PnP tests require.
- Consistency - ensure the cases across different languages (JavaScript/Python/C++) have the same functionality.
- Stability - whether we can obtain the statistic data stably.

You can `cherry-pick` these commits and run locally to see the result, thanks!

qiuzhong:
Got it.

minggangw:
The data we want to grab through the PnP tests are:

- CPU Cost
- Execution time
- Memory Cost

benchmark/rclnodejs/client-endurance-test.js
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('../../index.js');
+
+rclnodejs.init().then(() => {
+  console.log('The client will send a SetBool request every 100ms in 24 hours.');
+  console.log('Beginning');
```
Honry: As a tester, I would like to know when test will end, it would be nice if we could provide log of test beginning time/estimated end time.

benchmark/rclnodejs/client-endurance-test.js
```diff
@@ -0,0 +1,40 @@
...
+  let count = 0;
+
+  let timer = setInterval(() => {
+    if (count > 864000) {
+      clearInterval(timer);
+      rclnodejs.shutdown();
+      console.log('Done');
+    } else {
+      client.sendRequest(true, (response) => {
+        count++;
```
Honry: When the `response` could not been received in 100ms, `setInterval` will take priority to go into next loop, which may result in the overall execution time exceeds 24 hours.

benchmark/rclnodejs/client-endurance-test.js
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2018 Intel Corporation. All rights reserved.
```
Honry, line 1: I think a README.md for guiding tester to execute these PnP tests is necessary.

benchmark/rclnodejs/client-stress-test.js
```diff
@@ -0,0 +1,41 @@
...
+rclnodejs.init().then(() => {
+  console.log('The client will send a GetMap request(response contains a size of 10MB array) every 100ms in 1 hour.');
+  console.log('Beginning');
+
+  const node = rclnodejs.createNode('stress_client_rclnodejs');
+  const client = node.createClient('nav_msgs/srv/GetMap', 'get_map');
+  let count = 0;
+
+  let timer = setInterval(() => {
+    if (count > 72000) {
```
Honry: 72000 / 10 / 3600 = 2 hours? But it's 1 hour in the test desc.

minggangw: I will correct these problems, and add a README to demonstrate how to build and run the tests of Node.js/Python/C++, thanks!

minggangw:
@Honry @qiuzhong I have updated this PR per comments, please review it again. I will submit a new PR to add the README file.

benchmark/rclnodejs/client-stress-test.js
```diff
@@ -0,0 +1,45 @@
...
+  let sentTimes = 0;
+  let receivedTimes = 0;
+  let totalTimes = 36000;
+
+  let timer = setInterval(() => {
+    if (sentTimes++ > totalTimes) {
+      clearInterval(timer);
+    } else {
+      client.sendRequest({_dummy: true}, (response) => {
+        if (++receivedTimes > totalTimes) {
```
Honry: Is there a case that the `response` may be lost in some low performance device? Thus the `receivedTimes` could be never greater than `totalTimes`, i.e. the following code will never be called.

minggangw: If the network is broken, the client will not receive the response any more. 
If the device is low performance and the connection is still alive, I think the client will receive the response with high latency as the DDS is over TCP connection in ROS 2.

How about ignoring the count of sending the request and counting the response received only?

```JavaScript
 let timer = setInterval(() => {
      client.sendRequest({_dummy: true}, (response) => {
        if (++receivedTimes > totalTimes) {
          clearInterval(timer);
          rclnodejs.shutdown();
          console.log(`End at ${new Date().toString()}`);
        }
      });
}, 100);
```

PR #294, opened 2018-03-15
[PnP]Add PnP tests for rclnodejs client (Python)

### LINKED ISSUE #278: Add rclpy benchmark test used for PnP

minggangw:
This patch added the tests used for rclnodejs. We defined two kinds of
PnP tests, which are stress testing and endurance testing.

1. Stress testing
The test case is designed to find the behaviour beyond the normal
expected overload, and we can find the breaking points and short term
memory leaks through the tests. There is one test case for each of the
publisher/subscription and client/service. The strategy is:

- A publisher publishes a topic of std_msgs/UInt8MultiArray type every
  100ms, which contains an array of size 10MB, and sends 36000 times
  totally.
- A client sends a request of nav_msgs/GetMap whose response will
  contains an array of size 10MB until it receives the response from the
  service 36000 times.

2. Endurance testing
The test case is designed to find the expected overload for a longer
term, and we can find the long term memory leaks through the tests.
There is one test for each of the publisher/subscription and
client/service. The stragegy is:

- A publisher publishes a topic of std_msgs/JointState type every
  100ms for 24 hours (864000 totally).
- A client sends a request of std_srvs/SetBool until it receives the
  response from the service 864000 times.

All tests begin with a line of output 'Begin' and end with 'End'.

Fix #278

### DIFF
--- a/benchmark/rclpy/client-stress-test.py
+++ b/benchmark/rclpy/client-stress-test.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+# Copyright (c) 2018 Intel Corporation. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+from datetime import datetime
+from nav_msgs.srv import *
+
+def main():
+  rclpy.init()
+
+  print(
+    'The client will send a GetMap request continuously(response contains a size of 10MB array) until receiving 36000 response times.')
+  print('Begin at ' + str(datetime.now()))
+  node = rclpy.create_node('stress_client_rclpy')
+  client = node.create_client(GetMap, 'get_map')
+  request = GetMap.Request()
+  totalTimes = 36000
+  receivedTimes = 0
+
+  while rclpy.ok():
+    if receivedTimes > totalTimes:
+      node.destroy_node()
+      rclpy.shutdown()
+      print('End at ' + str(datetime.now()))
+    else:
+      client.call(request)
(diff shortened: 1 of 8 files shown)


benchmark/rclpy/client-stress-test.py
```diff
@@ -0,0 +1,44 @@
...
+
+  print('The client will send a GetMap request(response contains a size of 10MB array) every 100ms in 1 hour.')
+  print('Beginning')
+  node = rclpy.create_node('stress_client_rclpy')
+  client = node.create_client(GetMap, 'get_map')
+  request = GetMap.Request()
+  count = 0
+
+  while rclpy.ok():
+    if count > 72000:
```
qiuzhong: Here the `count` is about 2 hours.

benchmark/rclpy/publisher-stress-test.py
```diff
@@ -0,0 +1,64 @@
...
+  multiArray.data = [x & 0xff for x in range(1024 * 1024 * 10)]
+
+  print('The publisher will publish a UInt8MultiArray topic(contains a size of 10MB array) every 100ms in 1 hour.')
+  print('Beginning')
+  node = rclpy.create_node('stress_publisher_rclpy')
+  publisher = node.create_publisher(UInt8MultiArray, 'stress_topic')
+  count = 0
+
+  while rclpy.ok():
+    if count > 72000:
```
qiuzhong: 2 hours as well.

benchmark/rclpy/publisher-stress-test.py
```diff
@@ -0,0 +1,64 @@
...
+      node.destroy_node()
+      rclpy.shutdown()
+      print('Done')
+    else:
```
qiuzhong: No `count` increasing in else branch!

benchmark/rclpy/publisher-endurance-test.py
```diff
@@ -0,0 +1,55 @@
...
+  msg.velocity = [2.0, 3.0]
+  msg.effort = [4.0, 5.0, 6.0]
+  count = 0
+
+  while rclpy.ok():
+    if count > 864000:
+      node.destroy_node()
+      rclpy.shutdown()
+      print('Done')
+    else:
```
qiuzhong: Forgot to `count` increasing here?

minggangw: Thanks for pointing out the errors, I will update this PR later.

PR #303, opened 2018-03-30
Lock the handles throughout a loop in child thread

### LINKED ISSUE #301: Thread race condition
As `rclnodejs` is the dependency of `ros2-web-bridge`, there is [thread race condition](https://github.com/RobotWebTools/rclnodejs/blob/develop/src/executor.cpp#L102) which may lead the `ros2-web-bridge` server crash. 

There are already some tricky issues in ros2-web-bridge, like [#67](https://github.com/RobotWebTools/ros2-web-bridge/issues/67), [#69](https://github.com/RobotWebTools/ros2-web-bridge/issues/69). And we suspect they're all caused by this thread race condition.

minggangw:
Currently we only locked the mutex when adding the handles to the
wait_set, and there is a race condition that the handles may be
destroyed during executing the rcl_wait() function.

Instead we must lock the mutex throughout a loop to ensure that all
handles are valid which means none of them is destroyed. The mutex can
be unlocked when we have cleared the handles from the wat_set.

Fix #301

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -92,81 +92,90 @@ void Executor::Run(void* arg) {
 
   try {
     rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
-    rcl_ret_t ret = rcl_wait_set_init(
-        &wait_set, 0, 2, 0, 0, 0, rcl_get_default_allocator());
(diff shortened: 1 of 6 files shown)


qiuzhong:
@minggangw , I tested locally and it worked for me. But I have no idea why the all the CI fails now.

minggangw:
I have submitted #304 to disable the two failed cases, as it was not stable in CI circumstance.

Issue #316, opened 2018-04-13
Fix the disabled test suite blocking issue


qiuzhong:
The blacklist.json contains two test suites that are disabled in CI. 2 test cases were always blocked and cause a CI checking failed.

We thought it's caught by the low-performance CI systems. But now it also happened locally when we build rclnodejs with the latest ROS2 package from CI. Though it's not reproduced when rclnodejs was built with an old ROS2 package.

Should be fixed before 0.4.0 version release.

qiuzhong:
It happened in the Node.js rcl service with Python rcl client. In this condition, the workflow is like this:
* Launch Node.js rcl service and wait any other rcl client request
* Launch Python rcl client and it send a request
* Node.js rcl service handles this request and send a response

So far these steps all work. But it gets wrong at the next step:
* Python rcl client hangs after send request and doesn't receive the response.

I'm investing this.

qiuzhong:
This blocking is caused by the old sync `client.call(request)`, now it should use async way:

```
future = client.call_async(request)
rclpy.spin_until_future_complete(node, future)
```

Issue #313, opened 2018-04-12
v0.4.0 release


minggangw:
With the landing of [logging utility](https://github.com/RobotWebTools/rclnodejs/issues/281) and [benchmark tests](https://github.com/RobotWebTools/rclnodejs/issues/280), I think it is the proper opportunity to release v0.4.0 which will be the first **stable** release of rclnodejs (we have had 8 alpha&beta releases).

Meanwhile, I hope we can illustrate and compare the performance of rclnodejs with others (rclcpp&rclpy) through the statistics of PnP tests.

@Honry @hanyia @qiuzhong please append issues or patches you want to add into this release.

- [x] Run full testing based on our current [cases](https://github.com/RobotWebTools/rclnodejs/tree/develop/test ) on Linux/Windows/macOS
- [x] Run full testing based on [ros2-web-bridge](https://github.com/RobotWebTools/ros2-web-bridge)
- [x] Check our current online [document](http://robotwebtools.org/rclnodejs/docs/index.html)
- [ ] Add the PnP test result into [README](https://github.com/RobotWebTools/rclnodejs/blob/develop/README.md)
- [x] Issues to be fixed #316 
- [x] Benchmark #361

Honry:
@minggangw, great achievement! I think we should do a full testing before release.

qiuzhong:
And this full testing should also contain the protocol tests in `ros2-web-bridge` as a crash bug was found in `rclnodejs` when running ros2-web-bridge tests.

minggangw:
I will create a check list to track things we want to do in this release. Please edit that list if you have any requirement.

minggangw:
[v0.4.0](https://github.com/RobotWebTools/rclnodejs/releases/tag/0.4.0) has been published, we will add the PnP result later.

PR #317, opened 2018-04-16
[Update]Use async calling of Python rcl client request

qiuzhong:
Due the to rcl client API changes, use the aysnc calling of Python
rcl client request.

Fix: #316


minggangw (commented):
Thanks for finding the root cause, LGTM

minggangw:
Please change the status of this issue in #313, since it has been fixed, thanks!

PR #319, opened 2018-04-19
Support to generate .msg from .action files

### LINKED ISSUE #314: Add the actions feature in rclnodejs
We want to support and expose the [actionlib](http://wiki.ros.org/actionlib) interfaces by leveraging the [actionlib_nodejs](https://github.com/RethinkRobotics-opensource/actionlib_nodejs), which is an implementation of actionlib on ROS and ROS2. A another issue https://github.com/RethinkRobotics-opensource/actionlib_nodejs/issues/2 is tracking this task on the repo.

minggangw:
As the action specification is defined using a .action file, we have to
generated the needed .msg files(6 messages in total) based on that.

This patch added the function that the rosidl-generator can create the
.msg files when scanning the .action.

This commit is part of the actions feature for ROS2 (#314)

Fix #314


qiuzhong:
Where can I get the example .action files and try to generate the msg files?

minggangw:
You can find the wiki for [actionlib](http://wiki.ros.org/actionlib) in ROS1, and the example is [here](https://github.com/gerkey/ros1_external_use/tree/master/ros1_actions).

This PR only implemented the creation of .msg files from .action, and you have to add the path of your own project to the variable `AMENT_PREFIX_PATH` in order to be found by the `generator`. After execution, you will find 7 `.msg` files. 

To verify it, you can check if the generated files are the same with the ones generated by catkin of ROS1 (you have to [install ros1](http://wiki.ros.org/lunar/Installation/Ubuntu) if you want to do it, as the ament build tool does not support it yet).

minggangw:
One more thing, this patch is part of the actions feature of ROS2 and the following patches will be implemented later.

qiuzhong:
@minggangw , I tried to apply this PR to the code and generate `.msg` files based on a `.action` file. It works! It did generate 7 `.msg` files based on one `.action` file. The content of generated `.msg` files is almost the same as when they're generated from ROS1. 

But there is one difference: the `head` field is `std_msgs/Header` type from rosidl-generator while it's simply `Header` type from ROS1. Are they the same thing?

My next question is will you submit another PR to generate the `.js` files for the `.msg` packages generated based on the `.action` file?

minggangw:
>But there is one difference: the head field is std_msgs/Header type from rosidl-generator while it's simply Header type from ROS1. Are they the same thing?

In ROS2, the format of the message referenced in `.msg` file is `<package_name>`/`<message_name>`, please see the [example](https://github.com/ros2/common_interfaces/blob/9c05e50349bfc294ee73176f556635352ed5ca8a/sensor_msgs/msg/JointState.msg#L20).

>My next question is will you submit another PR to generate the .js files for the .msg packages generated based on the .action file?

The generation of `.msg` from `.action` is always the first step during the whole progress. When the `.msg` has been created, these files will be treated as the normal `.msg` files, thus the `.js` files are also created automatically. Please check under `generated`.

qiuzhong:
@minggangw , thanks for the clarification about the message type.

I checked the `generated` directory in `rclnodejs-1`  and there're no `.js` files that're related to the `.action` interface. 

I think there should be two rounds to generate `.js` files: one round to translate `.action` to `.msg` files. For the second round, it then translate the new generated `.msg` to `.js` files.

With only round, the new generated files might not be considered during `os.walk()`.

minggangw:
The `.action` file will not trigger generation of `js` files directly, but the `.msg` files created from `.action` does. Please check the path `generated/ros1_actions` where the `.js` files locate (should be 7 files).

qiuzhong:
There is no directory `generated/ros1_actions` generated.

qiuzhong:
@minggangw , any news about this PR?

minggangw:
The PR is being tested locally and will submit soon, thanks!

minggangw:
@qiuzhong I have updated this pr, you can test the functionality based on it, thanks!

qiuzhong:
That's great! I'll check if this can work.

qiuzhong:
Verified and it worked!  With one `rclnodejs.init()` calling, the `.msg` files and then `.js` files are generated from the `.action` files.

PR #320, opened 2018-04-23
[rosidl]Move the translator to rosidl-generator module

minggangw:
Currently, the translator, which can adapt a plain
JavaScript object to an object of ROS message, resides in the lib/
folder. It seems reasonable to move it to rosidl-generator and be
wrapped by the message class instead of exposing its interfaces to
rclnodejs.

This patch implemented the function described above.

Fix NONE

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,6 @@ const path = require('path');
 const QoS = require('./lib/qos.js');
 const rclnodejs = require('bindings')('rclnodejs');
 const validator = require('./lib/validator.js');
-const translator = require('./lib/message_translator.js');
 
 function inherits(target, source) {
   let properties = Object.getOwnPropertyNames(source.prototype);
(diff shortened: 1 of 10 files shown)


minggangw:
failure due to missing of `local_setup.bash` file in the binary package downloaded from ci.ros2.org, will try later.

minggangw:
blocked by https://github.com/ros2/ci/issues/157

dirk-thomas:
If you are using the nightly archives you need to source `prefix.bash` instead of `local_setup.bash` now. That is still in the process of being rolled out and not documented in the wiki yet.

minggangw:
@dirk-thomas thanks, I will change the file name. Shall I try until the next nightly build package? I do not find `prefix.bash`.

dirk-thomas:
> Shall I try until the next nightly build package?

Yes, it will only be inside the upcoming archive.

dirk-thomas:
Just a heads up: during the testing time of `colcon` we decided to rename the scripts to have the same names as the ones generated by the existing ROS build tools. So the archives generated in the next nightly builds will have `local_setup.*` files again (instead of `prefix.*` files). Sorry for the inconvenience resulting from the temporary shuffle.

minggangw:
@dirk-thomas thanks for your kind reminder!

Issue #331, opened 2018-05-15
Test cases for generating .msg from .action files


qiuzhong:
Since the rclnodejs is about to supporting action feature, see #314 and #319, there should be auto test cases to cover this.

qiuzhong:
PR #337 was merged on `actionlib` branch and needs to merge to the master branch

minggangw:
Yes, the `actionlib` is a feature branch and will be merged back to the `develop` when all implementations are ready, thanks!

minggangw:
Close, as already merged with `actionlib` branch

PR #337, opened 2018-05-21
[UT]Add test cases for covering action feature

qiuzhong:
Add test cases to cover action feature: check if a customed .action
can be used for generating .msg files and then .js files. (+14)

Fix: #331


minggangw (approved):
LGTM

Issue #339, opened 2018-05-25
AMENT_PREFIX_PATH splitter


qiuzhong:
On Linux and macOS, the path splitter is `:` while it's `;` on Windows. If the evrionment variable `AMENT_PREFIX_PATH` contains multiple parts of path, it should consider all platforms.

The following test suites didn't consider all platforms when splitting the path:
* test/test-message-object.js
* test/test-multi-nodes.js
* test/test-cross-lang.js
* test/test-rosidl-message-generator.js

```
test/test-multi-nodes.js:96:      var cppPubPath = path.join(process.env['AMENT_PREFIX_PATH'], 'lib', 'demo_nodes_cpp', 'talker');
test/test-multi-nodes.js:112:      var cppServicePath = path.join(process.env['AMENT_PREFIX_PATH'],
test/test-message-object.js:33:    const installedPackagesRoot = process.env.AMENT_PREFIX_PATH.split(':');
test/test-message-object.js:134:    const installedPackagesRoot = process.env.AMENT_PREFIX_PATH.split(':');
test/test-cross-lang.js:39:      var cppTalkPath = path.join(process.env['AMENT_PREFIX_PATH'], 'lib', 'demo_nodes_cpp', 'talker');
test/test-cross-lang.js:153:      var cppServicePath = path.join(process.env['AMENT_PREFIX_PATH'],
test/test-rosidl-message-generator.js:33:    const installedPackagesRoot = process.env.AMENT_PREFIX_PATH.split(':');

```
Some (test-multi-nodes.js and test-cross-lang.js ) are fixed in branch `actionlib` but the rest still should be fixed.

PR #340, opened 2018-05-25
[UT]Update path splitter for all platforms

qiuzhong:
Split the path by different splitters of platforms: `;` for Windows
and `:` for Linux and macOS.

Fix: #339


qiuzhong:
@minggangw

minggangw (approved):
LGTM

PR #347, opened 2018-06-01
[UT]Fix the failure of resolving the ros2 demo path on Windows

qiuzhong:
On Windows, there are .exe extension for the ros demo app path.
However, `fs.existsSync` API make sure the extension is considered.
Thus there is failure of resolving the ros2 demo path in test/utils.js.
This patch fixed this failure on Windows.


qiuzhong:
To keep actionlib and develop branch parallel, close it.

PR #357, opened 2018-06-08
Pre-generate the JavaScript messages during installation by postinstall

### LINKED ISSUE #356: Use postinstall to generate the JavaScript messages
As the rclnodejs will generate the JavaScript messages based on the IDL files (.srv/.msg), the time consumption of first-time start up is usually over several seconds.

We want to move the generation of the messages to the stage of installing to accelerate the start up speed for the first time.

minggangw:
To accelerate the speed of first-time start up, we better pre-generate
the messages during installation. Thus we do not have to wait the work
to be done when the module is initialized.

Fix #356

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "install": "node-gyp rebuild",
     "docs": "cd docs && make",
     "test": "node ./scripts/compile_cpp.js && node ./scripts/run_test.js",
-    "lint": "eslint --max-warnings=0 index.js scripts lib example rosidl_gen rosidl_parser test benchmark/rclnodejs && node ./scripts/cpplint.js"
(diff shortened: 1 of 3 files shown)


qiuzhong:
Tested this patch:
* First time `npm install`, `generated` directory and message files are generated.
* Remove `generated` and run `npm install`, message files are generated again.
* Remove `generated` and calling `rclnodejs.init()`, message files are generated as well.

So I think this patch works.

qiuzhong:
@minggangw , I have another concern: if this patch is merged with `actionlib` feature, will it still work?

minggangw:
This patch will be merged into `develop` branch. When the actionlib feature gets finished, the `actionlib` branch will merged with `develop`. This function will not be affected, which means all .msg/.srv/.action will be processed during installation.

minggangw:
We have to wait the latest package for xenial to fix the error on ci https://github.com/ros2/ci/issues/187

PR #370, opened 2018-06-29
[actionlib]Implement the actionlib on ROS 2.0

minggangw:
The actionlib stack provides a standardized interface for interfacing
with preemptable tasks in the current ROS 1.0. By leveraging the
actionlib_nodejs module, we migrate this feature to ROS 2.0 platform and
expose it through rclnodejs client.

This patch implements:

- The interfaces which wraps the client/server functions.
- Expose and configure the actionlib module in rclnodejs.
- Examples to show the usage of actionlib.

To understand the whole picture of actionlib, please visit
http://wiki.ros.org/actionlib/DetailedDescription

Fix #314

PR #374, opened 2018-07-04
Add test cases for covering actionlib feature

qiuzhong:


qiuzhong:
Please don't merge this before #370 is merged.

minggangw:
Hi @qiuzhong as the `actionlib` has been merged into `develop`, would you please rebase this PR on `develop` branch and update it, thanks!

qiuzhong:
I'll update it soon.

scripts/generate_messages.js
```diff
@@ -14,6 +14,16 @@
...
+let rclnodejsTestMsgsPath = path.join(rootDir, 'test', 'rclnodejs_test_msgs');
+let actionPath = path.join(rootDir, 'test', 'ros1_actions');
+process.env.AMENT_PREFIX_PATH = process.env.AMENT_PREFIX_PATH + 
+  path.delimiter + rclnodejsTestMsgsPath +
+  path.delimiter + actionPath;
+console.log('AMENT_PREFIX_PATH:', process.env.AMENT_PREFIX_PATH);
+
 const generator = require('../rosidl_gen/index.js');
 
 console.log('Start to generate the JavaScript messages...');
```
minggangw: Please do not change this file, as `npm install` will execute it during installation. We'd better not change the `process.env.AMENT_PREFIX_PATH` variable, the paths added here are only for testing.

I guess you also have to make change to the `run_test.js`

scripts/compile_tests.js
```diff
@@ -99,4 +99,22 @@ if (!fs.existsSync(publisherPath) && !fs.existsSync(subscriptionPath) &&
...
+    child.spawn('sh',
+      ['-c', `cp -r ${path.join(rootDir, 'install', 'ros1_actions') + '/.'} ${process.env.AMENT_PREFIX_PATH}`]);
+  }
+});
+actionBuildProcess.stdout.on('data', (data) => {
+  console.log(`${data}`);
+});
+actionBuildProcess.stderr.on('data', (data) => {
+  console.log(`${data}`);
+});
```
minggangw: Would you please abstract and wrap a function which can build and install a existing packages. So the `rclnodejs_test_msgs` can share the function with actionlib.

test/test-actionlib.js
```diff
@@ -0,0 +1,341 @@
...
+      done();
+    });
+    as.start();
+
+    let ac = new rclnodejs.ActionClientInterface({
+      type: 'ros1_actions/msg/DoDishes',
+      actionServer: 'dishes4',
+      rclnodejs: rclnodejs
+    });
+    ac.sendGoal({ goal: {dishwasher_id: 4}});
```
minggangw, line 75: I saw the time out happened on CI when running this case, Could you try to execute `sendGoal` periodically OR set an interval to avoid the contention.

scripts/generate_messages.js
qiuzhong: @minggangw , the main consideration of changing the `AMENT_PREFIX_PATH` is to generate js files for package `rclnodejs_test_msgs` and `ros1_actions` before testing. If this change must be removed from `npm install` then I have to generate the js packages twice. To make the test work, it's better to move the change to `compile_app.js`.

scripts/compile_tests.js
qiuzhong: I'll do that.

test/test-actionlib.js
qiuzhong: I'll take a try.

scripts/generate_messages.js
minggangw: For the current implement, all the JavaScript messages generated from IDL files are going to be removed before the running of test, please check [here](https://github.com/RobotWebTools/rclnodejs/blob/develop/scripts/run_test.js#L25)

``` JavaScript
fs.remove(path.join(path.dirname(__dirname), 'generated'), (err)=> {
  if (!err) {
    let mocha = new Mocha();
const testDir = path.join(__dirname, '../test/');
```

So I think you could append the path to `AMENT_PREFIX_PATH` in line [23](https://github.com/RobotWebTools/rclnodejs/blob/develop/scripts/run_test.js#L23). The messages will be re-generated before the test begins.

qiuzhong: I noticed that. However, the `ros1_action` package is a little tricky: it needs two rounds. 1st round will generate JavaScript message files and `CMakeLists.txt`. 2nd round will build the native libraries based on the `CMakeLists.txt`.  Maybe I should submit the generated `CMakeLists.txt` just like `rclnodejs_test_msgs`?

minggangw: I see and I suggest you could have the following steps:

* rclnodejs.regenerateAll() to generate the JS messages compulsively (a bunch of `.msg` and `CMakeLists.txt` will be present).
* Compile the `ros1_action` project by `CMakeList.txt` generated in step 1 and copy the installation files to the path of ROS 2.
*  Run mocha

Would you please investigate it to check if it's feasible? Thanks!

qiuzhong: Sure! I think it should work and I'll try it.

minggangw:
I checked the error from circleci & travis-ci, it seems that the rclnodejs can not find the libraries compiled from the package (action). Please double check whether the libraries are situated at the wanted folder, thanks!

qiuzhong:
The problem is the copying actions of `ros1_actions` and `rclnodejs_test_msgs` to ros2 failed. I'm investigating this.

qiuzhong:
@minggangw , CI errors fixed.

minggangw:
Great, I will merge it when the CI checking passes. Thanks!

minggangw (approved):
LGTM, good work 👍

PR #377, opened 2018-07-13
fix new/delete mismatch

martins-mozeiko:
Memory allocated with `new` must be deallocated with `delete`, not `free`.


minggangw (approved):
@martins-mozeiko thanks for pointing out this mess and fixing it.

PR #482, opened 2019-04-18
Fixes create client from string

### LINKED ISSUE #481: Create Interfaces from Strings
From the way ros2web-bridge is made it looks like interfaces can be loaded using strings encoded according to the format : ``{std_msd | ... }/{msg | srv}/{String | ...}``.
It looks like a function was written in this direction in the interface loader but was never called. For instance [here](https://github.com/mwamp/rclnodejs/blob/aefdc7195f7978620693d42768282985218468c4/lib/node.js#L248) a condition checks for a string but calls the generic loadInterface instead of loadInterfaceByString.

By the way AFAIK these functionnalities are not referenced anywhere? It would be nice to have some kind of specification/documentation for this so we know in the future why such choices were made. As I see it this is part of the api rclnodejs exposes yes?

IMO the whole idea of formatting the string as above doesn't feel very ROSish or JSish but I am no expert.

mwamp:
node.createClient should work when it's first parameter is of type : "{package}/{ type (oneOf[ 'msg', 'srv', ...])}/{Message or service Name}/"

partly fixes #481

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -246,8 +246,10 @@ class Node {
    * @return {Client} - An instance of Client.
    */
   createClient(typeClass, serviceName, options) {
-    if (typeof (typeClass) === 'string' || typeof (typeClass) === 'object') {
-      typeClass = loader.loadInterface(typeClass);
+    if (typeof (typeClass) === 'string') {
+      typeClass = loader.loadInterfaceByString(typeClass);
+    }else if( typeof (typeClass) === 'object') {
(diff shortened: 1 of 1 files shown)


mwamp:
aefdc71 fixes the issue in one context (client creation) it may arise in other (publisher suscribers?)
It seems their might also be a wanted behavior in case an object is passed instead of a string? 

This should be specified and implemented if so. Otherwise it might be wise to scrap reference to this from the code so that people are not mislead.

PR #507, opened 2019-08-22
Verify rclnodejs on Dasing Patch 3

minggangw:
Fix #None


minggangw:
The error happens to appveyor due to the https://github.com/ros2/rosidl/issues/388, which has not been merged into Dashing 3 (not seen on the nightly build), so the tests pass on all platforms now.

Close this issue.

Issue #498, opened 2019-06-26
Features needed to implement - umbrella issue


minggangw:
As [ROS 2](https://github.com/ros2/ros2) iterates rapidly, actually, there are two stable releases per year and a series of patch releases. I want to emphasize the target of this Node.js client is that it has the comparative feature set as the python client - [rclpy](https://github.com/ros2/rclpy).

But there is still a gap with rclpy, and we want to eliminate it finally. Due to limited resource, I'd like to invite any **volunteer** to contribute to this project, so the developers can have an additional choice besides C++ and Python languages. I believe that the requirement of web technology in ROS 2 ecosystem is strong, and JavaScript/Node.js enable more developers to work on ROS project quickly because of the low threshold.

I have tried my best to maintain rclnodejs module can work on the latest ROS 2 stable release, e.g.  Dashing Diademata, as soon as possible. I hope there are more contributors to implement the absent features and I will offer any help to move this project forward. Your contributions are welcome!

I list the features needed to implement below, so we can record the status. If anyone wants to handle it, please just assign it to yourself to indicate the feature is being implemented. Thanks!

- [ ] QoS - API and implementation for Liveliness and Deadline event callbacks #492 (Medial)
- [x] Actions support in rclnodejs #469 (High)
- [x] Support for parameters #416 (Medial)
- [x] Parameters can be passed through CLI #421 (Low)
- [x] Publish parameter events #415 (Medial)
- [x] Implement node parameters and parameter services #414 (Medial)
- [x] Add guard condition feature #315 (Medial)
- [x] Expose rcl_get_node_names #555
- [x] Add countPublishers and countSubscribers to Node #559
- [x] Support spin once #562 
- [x] Rate support #564
- [x] LifecycleNode support in rclnodejs #589  
- [x] Add getLogger to Node object #591 
- [x] Add getClock to Node object #592
- [x] Access current time from Node  #598 
- [x] Recognize the use_sim_time parameter #599 
- [x] Support to publish/subscribe serialized messages #646
- [x] Enable rclnodejs to fully support rcl contexts #668

P.S. 
I add an **indicator** (Low/Medial/High) to represent the combination of workload and difficulty of each issue. You can reference it to decide which feature will be suitable for you to handle.

Some suggestions in my mind:
- Please follow the [Google C++ coding style](https://google.github.io/styleguide/cppguide.html) to write your code, you can run `npm run lint` to verify it locally.
[...]

PR #509, opened 2019-09-11
Add pollForService feature

koonpeng:
I notice rclnodejs blocks indefinitely when doing a service call immediately after creating a client. This is possibly because the service is not ready yet.

This PR adds `isServiceServerAvailable` and `pollForService` method. They aim to provide the same functionality as `wait_for_service` in rclpy. (https://github.com/ros2/rclpy/blob/1cf216dd9ecd4359dc93f317e01b7386d5a36c5a/rclpy/rclpy/client.py#L146)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ project. The next step is to send a pull request (PR) for review. The PR will be
 Special thanks to the people who contribute.
 
 * [martins-mozeiko](https://github.com/martins-mozeiko)
+* [Teo Koon Peng](https://github.com/koonpeng)
 
 ## License
 
--- a/example/client-example.js
+++ b/example/client-example.js
@@ -25,7 +25,7 @@ rclnodejs.init().then(() => {
     b: Math.floor(Math.random() * 100),
   };
 
-  client.pollForService(1000).then(result => {
+  client.waitForService(1000).then(result => {
     if (!result) {
       console.log('Error: service not available');
       rclnodejs.shutdown();
--- a/example/client-example.js
+++ b/example/client-example.js
@@ -25,10 +25,17 @@ rclnodejs.init().then(() => {
     b: Math.floor(Math.random() * 100),
   };
 
-  console.log(`Sending: ${typeof request}`, request);
-  client.sendRequest(request, (response) => {
-    console.log(`Result: ${typeof response}`, response);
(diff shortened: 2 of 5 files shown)


minggangw:
Thanks for your contribution! Indeed, we should have these two functions to facilitate developers,  otherwise, they will meet the situation you describe.  Some suggestions:

- Would you please change the method name ```pollForService``` to ```waitForService```, as I prefer to align with `rclpy` naming to reduce confusion.
- About the CIs
  - For Travis-Ci (Unbuntu), there is a problem when installing the dependency base on the latest nightly build, but we can ignore it.
  - For the circleci (macOS), I notice some tests failed. I'm afraid we have to solve this before merging.
- Finally, you could add yourself to the [contributors](https://github.com/RobotWebTools/rclnodejs#contributors) 😄

koonpeng:
Thanks for the comments!

I changed `pollForService` to `waitForService` as suggested. I wan't sure what to name it because `pollForService` gives a better description but I can see how having the same name as rclpy helps users.

For the mac tests, it looks like it fails in `ActionClient status succeed` which this PR shouldn't affect. The 2nd failure seems to log that the publisher is no longer valid. Unfortunately I don't have a mac and I can't reproduce it on Linux. A subsequent run does pass though so it could be a one off error?

minggangw:
> For the mac tests, it looks like it fails in `ActionClient status succeed` which this PR shouldn't affect. The 2nd failure seems to log that the publisher is no longer valid. Unfortunately I don't have a mac and I can't reproduce it on Linux. A subsequent run does pass though so it could be a one off error?

Yes, some tests become flaky at a low rate, so you have to rerun it manually. Once it becomes green and we can land it.

The problem on travis-ci should have been fixed by https://github.com/ros2/ament_cmake_ros/issues/5, you can try to trigger the bot again after there is a newly built package on the ci.ros2.org. Thanks!

minggangw:
As all bots turn green, I create an issue #512 to track this feature and will merge it soon, thanks!

minggangw:
@koonpeng PR has been merged, thanks!

You are always welcome to contribute any new features to this module if you are interested in. Actually, #498 has a detailed summary of the current future absence compared with `rclpy`  and I hope we can offer an alternative to other developers to bridge them into ROS2 😃

Issue #511, opened 2019-09-13
maximum recursion error when running in jest


koonpeng:
There is a bug in `ref` which throws maximum recursion error when the module is loaded multiple times. Normally this error wouldn't happen because of the nodejs require cache. However this breaks rclnodejs when running in jest because jest ignores the require cache and ref is loaded multiple times, triggering the bug.

It can be reproduced with

a.test.js:
```
const rclnodejs = require('rclnodejs');

it('a', async () => {
  await rclnodejs.init();
  rclnodejs.shutdown();
});
```
b.test.js:
```
const rclnodejs = require('rclnodejs');

it('b', async () => {
  await rclnodejs.init();
  rclnodejs.shutdown();
});
```
`npx jest <test-dir>`

running the test, I got the error
```
 PASS  test/a.test.js
 FAIL  test/b.test.js
  ● b

    RangeError: Maximum call stack size exceeded

      at debug (../node_modules/ref/node_modules/debug/src/debug.js:1:1)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:746:3)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)
      at Object.writePointer [as _writePointer] (../node_modules/ref/lib/ref.js:747:11)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   0 total
Time:        0.365s, estimated 6s
Ran all test suites matching /.\/test/i.
```

There is a PR already submitted to ref that fix this but it hasn't been merged for a long time and ref doesn't seem to be maintained anymore.

minggangw:
Thanks for your findings! Through your description, I understand this problem happens only under the jtest framework, right? 

[...]

PR #514, opened 2019-09-16
add a note explaining a possible error that can happen when using jest

koonpeng:
add a note explaining the cause of #511


minggangw:
LGTM, thanks for the patch!

PR #523, opened 2019-10-14
Fix macOS platform error

minggangw:


minggangw:
The  [spdlog](https://github.com/ros2/spdlog_vendor) should have been included in the latest ROS packages, but it turns out the packages of macOS and Windows don't. I submitted a question (https://github.com/ros2/ros2/commits/master#commitcomment-35498489)  to ask the status and will change it accordingly later.

PR #527, opened 2019-10-25
Upgrade Node.js to LTS Erbium

minggangw:
Fix #None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,8 @@ jobs:
     - run: wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | bash
     - run: echo "source $HOME/.bashrc" >> ~/.bash_profile
     - run: cat ~/.bash_profile
-    - run: nvm install v10.19.0
-    - run: nvm alias default v10.19.0
+    - run: nvm install v12.16.1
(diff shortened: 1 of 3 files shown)


minggangw:
Currently, we cannot get passed on the latest Erbium which is v12.13.0 due to the incompatibility of [ref](https://www.npmjs.com/package/ref). So if we want to upgrade the Node.js in the future, we have to cope with this problem.

minggangw:
We also update the README to illustrate this, see 90ebf88b1298258c33ed290c5c2a030b0be9b312

minggangw:
I found some alternatives which support N-API, please check them out below:

- [ref-napi](https://github.com/node-ffi-napi/ref-napi)
- [ref-array-di](https://github.com/node-ffi-napi/ref-array-di)
- [ref-struct-di](https://github.com/node-ffi-napi/ref-struct-di)

I think we can replace the old `ref` series if we want to support a higher version of Node.js, e.g. Erbium

mattrichard:
@minggangw Once #582 goes in we should see if the CI builds pass in this PR.

minggangw:
Thanks, I am going to rebase and run the CIs again when it's landed.

minggangw:
Everything goes well! Thanks for fixing all the errors @mattrichard

PR #542, opened 2019-12-05
Disable the generation of actionlib

### LINKED ISSUE #510: build fail on ubuntu with dashing
When generating javascript messages, I got the following error message
```
Start to generate the JavaScript messages...
Caught error: Error: EACCES: permission denied, mkdir '/opt/ros/dashing/share/tf2_msgs/action/msg'
```
It looks like the current method to support action creates the `.msg` files from the `.action` files and uses them to generate the messages. This fails in ubuntu as its trying to create `.msg` files in a directory owned by root.

The result is that some messages are not generated, particularly packages that include both `.action` and `.msg` files, e.g. `example_interfaces`. This also causes some tests to fail as they rely on `example_interfaces`.

minggangw:
Fix #510

### DIFF
--- a/rosidl_gen/generator.json
+++ b/rosidl_gen/generator.json
@@ -1,6 +1,6 @@
 {
   "name": "rosidl-generator",
-  "version": "0.3.1",
+  "version": "0.3.2",
   "description": "Generate JavaScript object from ROS IDL(.msg) files",
   "main": "index.js",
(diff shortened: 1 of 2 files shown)


minggangw:
I'd like to merge this PR in advance, although there is still a failure on macOS, which caused by the wrong version of poco.

Issue #537, opened 2019-11-19
support TypeScript


wayneparrott:
I recognize and appreciate the commitment to ES6 by the project team. But I primarily favor TypeScript over JavaScript for much of my JS development. To that I recently hacked together a preliminary set of TypeScript typed definition (*.d.ts) files for the rclnodejs api for my own use. If there is interest I would gladly contribute this working to the rclnodejs project and support them going forward as the api evolves. See https://www.npmjs.com/package/@wayneparrott/rclnodejs-types
Alternatively, https://github.com/DefinitelyTyped/DefinitelyTyped could be a destination for the typedefs but personally I find projects that include integrated TS support feel more inclusive for TS developers and I would like to coordinate such efforts through this project before going outside it.

minggangw:
Wow, fantastic work! I still remember when I presented `rclnodejs` on ROSCon2018 in Madrid last year, I was asked whether the `rclnodejs` supports TypeScript, I think we can say that we have a solution that can use TypeScript now :smile:  Some comments:

> but personally I find projects that include integrated TS support feel more inclusive for TS developers and I would like to coordinate such efforts through this project before going outside it.

I agree it's more convenient for developers if we could integrate the TypeScript support inside `rclnodejs`, we can land the TS here first. Some work I can think out

-  How to integrate the TS into the current project, I suppose developers can select between JS and TS after the change
- How about the unit test, are we going to add some tests designed for TS interfaces?
- The document is going to be updated, we have [on-line](http://robotwebtools.org/rclnodejs/docs/index.html) version and the README

Finally, I think we can use this issue to track the following progress if you like, thanks again!

wayneparrott:
> How to integrate the TS into the current project, I suppose developers can select between JS and TS after the change

Correct. 
To get started I recommend that we manually maintain ts declaration files (*.d.ts) separate from the js api in /lib. For example maintain them in a directory named "types" or something similar. Once setup it takes very little time to maintain as the api evolves.

*Initially I tried using a jsdoc->*.d.ts conversion utility but it failed miserably. I hacked up the current set of types in a couple of hours, about the same about of time I spent trying to get the utility to work correctly and create something usable.*

Additionally, modify package.json to include a "types" property pointing to ./types/index.d.ts and extend the eslint configuration to include typescript support. 

> How about the unit test, are we going to add some tests designed for TS interfaces?

[...]

PR #538, opened 2019-11-25
Verify 0.10.3 on latest Eloquent

minggangw:

Issue #34, opened 2017-07-04
Figure the potential issue behind an endurance test


kenny-y:
In an endurance test that lasts for days:
 - Node.js client publisher: still working after 16,058,500 messages published (`rcl_init` and `rcl_shutdown` between every 2 messages)
 - Node.js client subscription: stopped working after 1,629,680 messages (got stuck and can't take any new messages)
 - C++ client subscription (`examples_rclcpp_minimal_subscriber_lambda`): stopped working after 6,185,940 messages (got stuck and can't take any new messages)

Note: a new instance of subscription of C++ or Node.js can take new messages from the Node.js publisher.

kenny-y:
CC @minggangw

minggangw:
It sees that with the latest master branch of [ROS2.0](https://github.com/ros2/ros2/blob/master/ros2.repos), the stability is better now and more robust, so close this.

PR #543, opened 2019-12-06
Add typescript support, #537

wayneparrott:
This PR is the initial implementation of #537, adding basic TypeScript support. 
It includes:
* type declaration files: see types/*.d.ts
* dtslint tests, see test/types
* updates to both readme files with new sections for use with TypeScript
* updated package.json with "types" property that points to types/index.d.ts, integrated dtslint to run as part of the test script, and eslint includes the types/ folder.
* tsconfig.json with tsc compiler options
* rosgen_tsd/index.js - script that runs as part of generate_messages.js to update the types/interfaces.d.ts with type aliases


wayneparrott:
Bummer! Failed during the build phase:
 "Library not loaded: @rpath/librcl.dylib"

I successfully build and install on my Linux system with ros2 eloquent using the same node version, 10.16.1. 

Thoughts on how to move forward?

minggangw:
Hi thanks for your PR, that's something. For the error that happened on CircleCI, I think it not caused by your code, I suggest you could replace the [latest stable packages](https://ci.ros2.org/view/packaging/) of ROS2 with Eloquest packages. The reason why I choose the latest stable packages against `develop` branch is that we want to find the incompatibility with rcl library as early as possible, but the downside is that the package tends to be kind of unstable, e.g. the current failure on CIs.

So I suggest you could change the scripts used by the CIs to use Eloquest directly, please reference the PR https://github.com/RobotWebTools/rclnodejs/pull/538, meanwhile, please don't forget to add branch `typescript` to be built, e.g. [here](https://github.com/RobotWebTools/rclnodejs/blob/develop/.travis.yml#L8), otherwise, the bots will not be triggered. As your code only contains pure JavaScript, I deem that passing the Linux platform is enough to verify the functionality. Thanks!

wayneparrott:
Hitting an issue with the mac os build when running generate_messages.js a dll is not being found. I merged #538 to use eloquent release. Would appreciate any guidance you can provide for resolving this matter.

/Users/distiller/RobotWebTools/rclnodejs/node_modules/bindings/bindings.js:121
        throw e;
        ^

Error: dlopen(/Users/distiller/RobotWebTools/rclnodejs/build/Release/rclnodejs.node, 1): Library not loaded: @rpath/librcl.dylib
  Referenced from: /Users/distiller/RobotWebTools/rclnodejs/build/Release/rclnodejs.node
  Reason: image not found
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:807:18)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at bindings (/Users/distiller/RobotWebTools/rclnodejs/node_modules/bindings/bindings.js:112:48)
    at Object.<anonymous> (/Users/distiller/RobotWebTools/rclnodejs/lib/logging.js:18:38)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! rclnodejs@0.10.4 postinstall: `node scripts/generate_messages.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the rclnodejs@0.10.4 postinstall script.

minggangw:
>  1): Library not loaded: @rpath/librcl.dylib

It seems that the library cannot be found. It's weird that it passed with #538 with the same configuration. I am going to check it, don't worry.

rostsd_gen/index.js
```diff
@@ -0,0 +1,506 @@
...
+		}
+	}
+}
+*/
+
+'use strict';
+
+const path = require('path');
+const fs = require('fs');
+const rclnodejs = require('../index.js');
```
minggangw: I suspect this line causes the runtime error when loading the dynamic library. I haven't investigated why this only happens on macOS, probably we executed:

``` npm install --python=python2.7```

**Comment by wayneparrott** [comment]
Making some progress now with your previous hint re: loading rclnodejs.js. That was the root cause of the problem with not being able to locate a dll on mac osx. I removed the import of the rclnodejs.js module and success. Thanks for the help. 

While moving forward, still not passing CI tests due to 2 test fails. I will try to look into them later today as they appear to be basic test failures and not related to any of the typescript-based changes.

  1) Cross-language interaction
       Node.js publisher
         Cpp subscription should receive msg from Node.js publisher:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/distiller/RobotWebTools/rclnodejs/test/test-cross-lang.js)
  

  2) Multiple nodes interation testing
       Publisher/Subscription
         Node.js publisher - Python and Cpp subscription:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/distiller/RobotWebTools/rclnodejs/test/test-multi-nodes.js)

**Comment by minggangw** [comment]
The tests targeted at interaction with other clients (Python/C++) become flaky sometimes, this happens on the CI machines, not locally. But I still like to retain them because they can verify the compatibility with other clients (some potential changes may occur). I am going to re-run the bots again, hopefully it can pass this time :smile:

**Review by minggangw — COMMENTED** [review]
I think you submitted the binary file `test/cpp/listener` which is for linux caused the error, you may add it mistakenly. Would you please remove it and run again?

**Comment by wayneparrott** [comment]
Inching towards passing CI. :)
Thanks for the tip on the listener exe which I accidentally checked in. 
Now if we can get the Windows platform CI to pass.

**Comment by minggangw** [comment]
Oops, this error happens on `develop` branch constantly as well, which reports:

``` bash
failed to load shared library of rmw implementation. Exception: Cannot load library: c:\ros2-windows\bin/rmw_fastrtps_cpp.dll
```

minggangw:
The problem should be fixed by #544, please rebase your PR and have a another run :smile:

wayneparrott:
**WOOT!** CI finally passed.
Thanks for all of your assistance. Future PRs will go much more smoothly as I think I have the process down.

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - run: brew install openssl
     - run: brew install asio console_bridge
     - run: python3 -m pip install catkin_pkg empy git+https://github.com/lark-parser/lark.git@0.7d pyparsing pyyaml setuptools argcomplete colcon-common-extensions numpy
-    - run: mkdir -p ~/ros2_install && cd ~/ros2_install && wget https://ci.ros2.org/view/packaging/job/packaging_osx/lastStableBuild/artifact/ws/ros2-package-osx-x86_64.tar.bz2 && tar xf ros2-package-osx-x86_64.tar.bz2
+    - run: mkdir -p ~/ros2_install && cd ~/ros2_install && wget https://github.com/ros2/ros2/releases/download/release-eloquent-20191122/ros2-eloquent-20191122-macos-amd64.tar.bz2 && tar xf ros2-eloquent-20191122-macos-amd64.tar.bz2
```
minggangw, line 27: We may change back to the nightly build when we merge back to the `developer` branch.

.travis.yml
```diff
@@ -5,9 +5,7 @@ sudo: required
 
 branches:
   only:
-  - develop
-  - master
-  - actionlib
+  - typescript
```
minggangw, line 8: We may add `developer` branch later

package.json
```diff
@@ -1,20 +1,22 @@
 {
   "name": "rclnodejs",
-  "version": "0.10.3",
+  "version": "0.10.4",
```
minggangw: I think we can bump to 0.11.0, as we offer a new feature from this version.

rostsd_gen/index.js
```diff
@@ -0,0 +1,505 @@
...
+// scan generated files, collect pkg and msg info
+function getPkgInfos(generatedRoot) {
+  let pkgInfos = [];
+  const rootDir = generatedRoot;
+  let pkgs = fs.readdirSync(rootDir);
+
+  for (let pkg of pkgs) {
+
+    if (pkg.endsWith('.json')) continue;
+    //if (pkg.startsWith('actionlib')) continue;
```
minggangw: Could this line be deleted?

rostsd_gen/index.js
```diff
@@ -0,0 +1,505 @@
...
+
+    const pkgPath = path.join(rootDir, pkg);
+    const files = fs.readdirSync(pkgPath);
+
+    for (let filename of files) {
+      const typeClass = fileName2Typeclass(filename);
+
+      if (typeClass.type === 'srv') { // skip __srv__<action>
+        if (!typeClass.name.endsWith('Request') && !typeClass.name.endsWith('Response')) {
+          //console.log('skip:', typeClass.name);
```
minggangw: Ditto

rostsd_gen/index.js
```diff
@@ -0,0 +1,505 @@
...
+`;
+    fs.writeSync(fd, pkgNamespaceTemplate);
+
+    let curNS = null;
+    for (const msgInfo of pkgInfo.messages) {
+
+      // omit namespaces that are not 'msg' or 'srv'
+      // e.g., action_tutorials.action.FibonacciActionXXX
+      //if (msgInfo.typeClass.type.startsWith('action')) continue;
+      // if (msgInfo.typeClass.name.includes('Fibonacci')) continue;
```
minggangw: Are these 4 lines still meaningful?

rostsd_gen/index.js
```diff
@@ -0,0 +1,505 @@
...
+    new typeClass() :
+    undefined;
+}
+
+
+const tsd_generator = {
+  // version() {
+  //   // eslint-disable-next-line
+  //   return fse.readJsonSync(path.join(__dirname, 'generator.json')).version;
+  // },
```
minggangw: I think we could delete the comment if we are not going to expose it.

rostsd_gen/index.js
```diff
@@ -0,0 +1,505 @@
...
+  ],
+  "baseType": {
+    "pkgName": "sensor_msgs",
+    "type": "LaserScan",
+    "stringUpperBound": null,
+    "isPrimitiveType": false
+  },
+  "msgName": "LaserScan"
+}
+*/
```
minggangw: Could we delete this note?

types/interfaces.d.ts
```diff
@@ -0,0 +1,1796 @@
...
+    visualization_msgs.msg.InteractiveMarkerFeedback |
+    visualization_msgs.msg.InteractiveMarkerInit |
+    visualization_msgs.msg.InteractiveMarkerPose |
+    visualization_msgs.msg.InteractiveMarkerUpdate |
+    visualization_msgs.msg.Marker |
+    visualization_msgs.msg.MarkerArray |
+    visualization_msgs.msg.MenuEntry |
+    visualization_msgs.srv.GetInteractiveMarkers_Request |
+    visualization_msgs.srv.GetInteractiveMarkers_Response;
+}
```
minggangw: I have a question here, what's going to happen if a user creates a `.msg`/`.srv` defined by himself? how can offer the type information?

minggangw (approved):
Thanks for the colossal PR and add this great feature! Some nit comments.

.circleci/config.yml
wayneparrott: +1

.travis.yml
wayneparrott: +1

package.json
wayneparrott: OK. I'll make this change. to 0.11.0

rostsd_gen/index.js
wayneparrott: +1, I'll make this change.

rostsd_gen/index.js
wayneparrott: +1, I'll remove the commented out line.

rostsd_gen/index.js
wayneparrott: +1, will remove.

rostsd_gen/index.js
wayneparrott: +1, I'll remove this code.

rostsd_gen/index.js
wayneparrott: +1, I'll remove the tmp developer notes.

types/interfaces.d.ts
wayneparrott: My understanding is that when a user creates a custom msg or srv, he must run rclnodejs.regenerateAll() to generate the new js structures. But unfortunately I did not account for regenerateAll() to additionally generate a new interfaces.d.ts updated with new type info for the new msg or srv. My BAD as I only modified the scripts/generate_messages.js#generateAll() to generate the interface.d.ts file which is only run at installation time.

To fix this missing functionality my plan is to update the regenerateAll() function to also generate the interface.d.ts file. Do you agree?

Alternatively, I can add the logic for generating the interface.d.ts file as part of the rosidl_gen/index.js#generateAll() function. 

Thoughts? 

I'll make the changes and test tomorrow.

rostsd_gen/index.js
wayneparrott: You were correct. I removed line #34 which resolved this problem. Thx!

types/interfaces.d.ts
minggangw: > To fix this missing functionality my plan is to update the regenerateAll() function to also generate the interface.d.ts file. Do you agree?

I agree we should update the generated interface when any change happens. What's more, as the `types/interfaces.d.ts` is generated dynamically, we'd better not to submit this into repo and it only exists locally. Thanks!

rostsd_gen/index.js
wayneparrott: fixed.

rostsd_gen/index.js
wayneparrott: fixed.

rostsd_gen/index.js
wayneparrott: fixed.

rostsd_gen/index.js
wayneparrott: fixed.

rostsd_gen/index.js
wayneparrott: fixed.

types/interfaces.d.ts
wayneparrott: fixed. See index.js#regenerateAll() which calls the tsd generator after generating the javascript code.

PR #557, opened 2020-01-31
Fix 556: added missing #include stmt

wayneparrott:
Added missing #include statement.

To verify,
```
>cd rclnodejs
>node-gyp --debug rebuild
```


minggangw:
Let's ignore the failure on circleci due to the wrong version of poco, merging...

PR #559, opened 2020-02-10
Add countPublishers and countSubscribers to Node

mattrichard:
Adds countPublishers and countSubscribers which was missing from the Node class. rclpy implementation can be found [here](https://github.com/ros2/rclpy/blob/196669e539dd51bc56f9f4fdf6f0222d99480d0d/rclpy/rclpy/node.py#L1517) for comparison.

Fix #None

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -462,6 +462,30 @@ class Node {
   getNodeNames() {
     return rclnodejs.getNodeNames(this.handle);
   }
+
+  /**
+   * Return the number of publishers on a given topic.
+   * @param {string} topic - The name of the topic.
+   * @returns {number} - Number of publishers on the given topic.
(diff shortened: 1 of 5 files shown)


mattrichard:
Not sure why AppVeyor keeps timing out. Does this happen a lot? Tests work fine locally on both Linux and Windows.

image

minggangw:
The ut running on AppVeyor seems less stable than the local one, sometimes, the tests, which aim to verify the interaction between rclnodejs client and rclpy/rclcpp clients, tend to cause the failures (usually timeout). I am going to re-run it again.

minggangw (approved):
Thanks for implementing the missing interfaces and the CIs have passed, will merge soon!

PR #561, opened 2020-02-13
Add support for guard conditions

### LINKED ISSUE #315: Add guard condition feature
We should implement the guard condition based on [rcl](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/guard_condition.h#L32) library.

mattrichard:
Implements guard conditions in rclnodejs. In addition, now only entities that are ready to be handled will be executed. This was initially needed since guard conditions should only run when triggered, but was extended to all entities. The approach taken was passing all RclHandles that are ready to be executed back to the main process.

All changes to the .cpp files should be carefully reviewed for quality, thread safety, and alignment with future needs. There's a lot of looping occurring now, but this is not much different than the  rclpy implementation.

Fix #315

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,6 +118,7 @@ let rcl = {
     node.init(nodeName, namespace);
     debug('Finish initializing node, name = %s and namespace = %s.', nodeName, namespace);
     node.handle = handle;
+    node.context = context;
     this._nodes.push(node);
     return node;
   },
--- a/lib/guard_condition.js
+++ b/lib/guard_condition.js
@@ -1,0 +1,50 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('bindings')('rclnodejs');
+const Entity = require('./entity.js');
+const Context = require('./context.js');
+
+/**
+ * @class - Class representing a guard condition in ROS
+ * @hideconstructor
+ */
+
+class GuardCondition extends Entity {
+  constructor(handle, callback) {
+    super(handle, null, null);
+
+    this._callback = callback;
+
+    // True when the executor sees this has been triggered but has not yet been handled
+    this._executorTriggered = false;
+  }
+
+  get callback() {
+    return this._callback;
+  }
+
+  static createGuardCondition(callback, context = Context.defaultContext()) {
+    let handle = rclnodejs.createGuardCondition(context.handle());
+    return new GuardCondition(handle, callback);
+  }
+
+  trigger() {
+    rclnodejs.triggerGuardCondition(this.handle);
+  }
+}
+
+module.exports = GuardCondition;
--- a/lib/guard_condition.js
+++ b/lib/guard_condition.js
@@ -1,5 +1,3 @@
-// Copyright (c) 2017 Intel Corporation. All rights reserved.
-//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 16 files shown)


lib/guard_condition.js
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
+//
```
minggangw: These 2 lines are needed when Intel employees contribute codes to open source projects and this project is not managed by Intel, so I think you can remove them.

index.js
```diff
@@ -118,6 +118,7 @@ let rcl = {
     node.init(nodeName, namespace);
     debug('Finish initializing node, name = %s and namespace = %s.', nodeName, namespace);
     node.handle = handle;
+    node.context = context;
```
minggangw, line 121: Thanks for fixing this.

lib/guard_condition.js
```diff
@@ -0,0 +1,50 @@
...
+  get callback() {
+    return this._callback;
+  }
+
+  static createGuardCondition(callback, context = Context.defaultContext()) {
+    let handle = rclnodejs.createGuardCondition(context.handle());
+    return new GuardCondition(handle, callback);
+  }
+
+  trigger() {
```
minggangw, line 44: Please add the comment block for jsdoc

test/test-guard-condition.js
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
+//
```
minggangw: ditto

test/test-guard-condition.js
```diff
@@ -0,0 +1,96 @@
...
+  afterEach(function() {
+    node.destroy();
+  });
+
+  it('Test trigger', async function() {
+    let called = false;
+
+    function func() {
+      called = true;
+    }
```
minggangw: nit: we could simplify by using [mocha-sinon](https://www.npmjs.com/package/mocha-sinon), e.g.

```JavaScript
let sinon = require("sinon");
let callback = sinon.spy()
...

// After has been triggered.
expect(callback).to.be.calledOnce;
```

lib/node.js
```diff
@@ -47,15 +49,32 @@ class Node {
...
+
+    // Retrigger a guard condition that was triggered but not handled
+    this._guards.forEach((guard) => {
+      if (guard._executorTriggered) {
+        guard.trigger();
+      }
+    });
+
+    // Mark all guards as triggered before processing any handlers since they're auto-taken
+    guardsReady.forEach((guard) => guard._executorTriggered = true);
```
minggangw: is it necessary here?

lib/node.js
```diff
@@ -47,15 +49,32 @@ class Node {
...
+    let timersReady = this._timers.filter((timer) => handles.indexOf(timer.handle) !== -1);
+    let guardsReady = this._guards.filter((guard) => handles.indexOf(guard.handle) !== -1);
+    let subscriptionsReady = this._subscriptions.filter((subscription) =>
+      handles.indexOf(subscription.handle) !== -1);
+    let clientsReady = this._clients.filter((client) => handles.indexOf(client.handle) !== -1);
+    let servicesReady = this._services.filter((service) => handles.indexOf(service.handle) !== -1);
+
+    // Retrigger a guard condition that was triggered but not handled
+    this._guards.forEach((guard) => {
+      if (guard._executorTriggered) {
```
minggangw: I look through the implementation of rclpy, and found that the execution of the callback of the guard and pushing the guard into |guardsReady| array are on different threads, but we do this in the main thread only, so it's sync. Maybe we can remove this, but I haven't tested locally. Correct me if I'm wrong.

minggangw (commented):
Thanks for making this change which reduces the overhead in each event loop. The second commit makes this PR perfect :) and some comments below, thanks!

test/test-guard-condition.js
mattrichard: Nice to know. That's much better. Seems like sinon alone will suffice for now.

lib/node.js
mattrichard: Did some further testing and is appears it is not needed.

lib/node.js
mattrichard: Looks like you're right. Did some further testing and this can be removed.

mattrichard:
@minggangw Assuming there are no further comments, do you want me to rebase or will you squash the merge instead?

minggangw:
@mattrichard it looks good to me and I think we have improved the performance by this PR not only the feature. I'm going to squash them soon, thanks!

PR #563, opened 2020-02-15
Implement rclnodejs.spinOnce

### LINKED ISSUE #562: Support spin once
rclnodejs has no support for spin once (that I'm aware of) like in rclcpp and rclpy. In #561 I had to resort to using setTimeout instead in the unit test to wait for the event loop to finish once. At a minimum, spin once should help improve unit tests within rclnodejs and make them more reliable.

mattrichard:
Adds spinOnce support to rclnodejs. Unlike spin which runs in a separate thread, spinOnce will run synchronously and block the caller until the event loop has finished or times-out. Ideally, this should result in more predictable execution of unit tests that rely on the event loop. spinOnce cannot be used if already spinning and will throw error if attempted.

Usage:

```javascript
let node = rclnodejs.createNode('my_node');
rclnodejs.spinOnce(node);
```

Fix #562

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -70,8 +70,10 @@ void Executor::SpinOnce(rcl_context_t* context, int32_t time_out) {
         ExecuteReadyHandles();
 
       if (rcl_wait_set_fini(&wait_set) != RCL_RET_OK) {
-        throw std::runtime_error(std::string("Failed to destroy guard waitset:") +
-                                rcl_get_error_string().str);
+        std::string error_message =
+            std::string("Failed to destroy guard waitset:") +
+            std::string(rcl_get_error_string().str);
+        throw std::runtime_error(error_message);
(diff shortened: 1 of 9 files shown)


src/executor.cpp
```diff
@@ -176,4 +152,64 @@ void Executor::Run(void* arg) {
...
+    std::string error_message =
+        std::string("Failed to clear wait set: ") +
+        std::string(rcl_get_error_string().str);
+    throw std::runtime_error(error_message);
+  }
+
+  return status != RCL_RET_WAIT_SET_EMPTY;
+}
+
+void Executor::ExecuteReadyHandles() {
```
minggangw, line 204: It's so nice to encapsulate these functions.

minggangw (approved):
Thanks for the PR, LGTM!

PR #565, opened 2020-02-16
Upgrade deep-equal and mkdirp

minggangw:
This patch upgrades deep-equal and mkdirp to the latest stable versions.

Fix #None

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,6 @@ const QoS = require('./lib/qos.js');
 const rclnodejs = require('bindings')('rclnodejs');
 const tsdGenerator = require('./rostsd_gen/index.js');
 const validator = require('./lib/validator.js');
-const ActionLib = require('ros2-actionlibjs');
 const Time = require('./lib/time.js');
 const TimeSource = require('./lib/time_source.js');
 const {Clock, ROSClock} = require('./lib/clock.js');
@@ -98,8 +97,6 @@ let rcl = {
   /** {@link Duration} class */
   Duration: Duration,
 
-  ActionLib: ActionLib,
-
   /**
    * Create a node.
    * @param {string} nodeName - The name used to register in ROS.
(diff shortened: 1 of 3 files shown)


minggangw:
Based on the latest `deep-equal` (2.0.1), the following returns `false`

```JavaScript
deepEqual([ 8, 9, 6, 4 ], Float64Array.from([ 8, 9, 6, 4 ]))
```
So we must figure it out, keep the current version or change the cases.

mattrichard:
> So we must figure it out, keep the current version or change the cases.

I'm always for keeping dependencies up-to-date. One thing though, do we really need the library? Does it's perf benefits make any noticeable difference?

minggangw:
> Does it's perf benefits make any noticeable difference?

Yes, it's said that "This module is around 46 times faster than wrapping assert.deepEqual() in a try/catch."

mattrichard:
This gets even worse in Node.js v12. The following will now return false.

```javascript
assert.deepEqual([ 8, 9, 6, 4 ], Float64Array.from([ 8, 9, 6, 4 ]))
```

Actually, this change to `assert.deepEqual` is actually the reason why deep-equal's behavior was updated. Also, assert.deepEqual is deprecated in v12 (in favor of deepStrictEqual).

Looks like we're going to have to rely on deep-equal more if we want to continue using non-strict comparisons. As part of nodejs v12 support we should replace all assert.deepEqual with deep-equal. It's up to you whether we should update deep-equal or not.

minggangw:
It seems that we only have one case that's not compatible with the latest `deep-equal`, which is the comparison of `Array` with `TypedArray`. How about offering a util method to handle it (compare the value by index), and then we can use the `deep-equal` for the others? The downside is we will lose some performance when running the test cases.

minggangw:
Please ignore my last comment, I misunderstood. I suppose if we want to carry out this kind of comparison we need to keep the `deep-equal` @1.1.1, otherwise, we have to customize our own `deepEqual`.

I'd like to move `deep-equal` to `devDependencies` and keep the current version.

mattrichard:
> I'd like to move deep-equal to devDependencies and keep the current version.

Sounds good to me. It'll make supporting v12 easier.

PR #566, opened 2020-02-16
Add getNodeNamesAndNamespaces to Node

mattrichard:
Fix getNodeNames return type and adds getNodeNamesAndNamespaces to align with the interface of Node in rclpy more closely.

This should go in before 0.12 release, as it would otherwise be a breaking change to the Node interface.

Fix #None


minggangw (approved):
LGTM and thanks for doing the improvement!

PR #567, opened 2020-02-16
Fix v8 deprecation warnings

mattrichard:
Fixes all v8 deprecations allowing us to target higher versions of Node.js. With this change Node.js versions <8 will no longer be supported.

The minimum version of node documented in the README as officially supported was updated to v8.12. Version 8.12 was chosen because it is the first version where N-API was considered stable. See [here](https://github.com/nodejs/abi-stable-node/blob/doc/README.md). Most versions prior to this gave me errors with int64-napi when running the unit tests on Windows.

Marked as draft as this shouldn't go out with v0.12 release and there will likely be more deprecated APIs used once parameter support is added.

Fix #None


minggangw:
Thanks for improving this and we can have a clean output during compiling now 👍

minggangw:
Shall we land it on the `developer` branch now?

minggangw (approved):
LGTM and merging, thanks!

PR #570, opened 2020-02-17
Add actions to the event loop

mattrichard:
Adds action clients and servers into the event loop. You can't add action client/servers right now so this change essentially does nothing.

Actions present an interesting problem where not only do you need to know if the handle is ready, but you also need to know what to execute within the handle. rclpy abstracts actions with a Waitable base class and lets the sub-class fill in the details of reading from the wait_set. This approach doesn't fit in with the design of rclnodejs, so instead actions are treated as entities, similar to publishers, subscribers, etc, but with some special handling.

To know what part of the action server/client should be processed a property bag has added to RclHandle allowing the event loop to attach arbitrary data to be sent back to the main thread. The property bag could be extended in the future if ever needed by other entity types.

Below is how the property bag would be used in later PRs.

```javascript
// From Node class
execute(handles) {
  let actionServersReady = this._actionServers.filter((actionServer) =>
      handles.indexOf(actionServer.handle) !== -1);

  // ...

  actionServersReady.forEach((actionServer) => {
    // Read the properties from the handle
    let properties = actionServer.handle.properties;

    if (properties.isGoalRequestReady) {
      // process goal request
    }

    if (properties.isCancelRequestReady) {
      // process cancel request
    }

    // process other requests
  }
}
```

Note this PR is targeting the 'actions' branch.

### DIFF
--- a/src/rcl_handle.hpp
+++ b/src/rcl_handle.hpp
@@ -43,7 +43,7 @@ class RclHandle : public Nan::ObjectWrap {
   void Reset();
   void AddChild(RclHandle* child) { children_.insert(child); }
   void RemoveChild(RclHandle* child) { children_.erase(child); }
-  void SetProperty(const std::string& name, bool value) {
+  void SetBoolProperty(const std::string& name, bool value) {
       properties_[name] = value; }
   void SyncProperties();
 
--- a/src/rcl_handle.hpp
+++ b/src/rcl_handle.hpp
@@ -19,6 +19,7 @@
 
 #include <functional>
 #include <set>
+#include <map>
 
 namespace rclnodejs {
 
@@ -42,6 +43,9 @@ class RclHandle : public Nan::ObjectWrap {
   void Reset();
   void AddChild(RclHandle* child) { children_.insert(child); }
   void RemoveChild(RclHandle* child) { children_.erase(child); }
+  void SetProperty(const std::string& name, bool value) {
+      properties_[name] = value; }
+  void SyncProperties();
 
  private:
   RclHandle();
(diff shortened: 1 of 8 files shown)


mattrichard:
This was tested by hacking together an action client/server and it worked great. I was able to successfully send and receive a goal request, but obviously the actual messages are all messed up right now since the code gen needs to be updated.

The big question with this PR is whether the addition of a property bag on the handles makes sense or if there are better approaches. Maybe it's time to start abstracting the wait_set handling more. I personally don't think any abstractions will be of much help as of right now, maybe some point in the future as the library expands.

src/rcl_handle.hpp
```diff
@@ -42,6 +43,9 @@ class RclHandle : public Nan::ObjectWrap {
   void Reset();
   void AddChild(RclHandle* child) { children_.insert(child); }
   void RemoveChild(RclHandle* child) { children_.erase(child); }
+  void SetProperty(const std::string& name, bool value) {
```
minggangw: how about `SetBoolProperty`, as `SetProperty` sounds too general.

minggangw:
Some limitations of Node.js make it not as powerful as Python when developing C++ addon, some I know:

- The main thread cannot be blocked
- We cannot manipulate threads explicitly by JavaScript (addons like [napajs](https://github.com/microsoft/napajs) can extend the capability)
- We cannot call methods of v8 objects in sub-thread which was created on the main thread originally  

So we have to take care of the inter communication between the main thread and sub-thread.

>The big question with this PR is whether the addition of a property bag on the handles makes sense or if there are better approaches. Maybe it's time to start abstracting the wait_set handling more. I personally don't think any abstractions will be of much help as of right now, maybe some point in the future as the library expands.

I agree that we can attach the properties to the RclHandle class at the current stage and when the action feature is done, we could refactor the current executor mechanism because we will have a concrete scope then.

mattrichard:
@minggangw If you have no further comments I'll merge this PR. Also, if you don't mind could you re-trigger the PR builds? I just updated the branch to include your commits that use the latest ROS build artifacts.

minggangw:
Please merge it by yourself, it's fine with me. I am going to trigger the CI, thanks!

PR #578, opened 2020-02-20
Replace ref libraries with v12 compatible alternates

mattrichard:
Replaces the ref* libraries with the alternatives mentioned in #527. These seem like our best options right now.

Fix #None


minggangw:
Thanks and merging...

PR #582, opened 2020-02-23
Fix compile errors with node v12

mattrichard:
Fixes all compiler errors with NodeJS v12. Also switches from using the deprecated assert.deepEqual to deep-equal in unit tests.


minggangw (approved):
Nice work and merging...

PR #583, opened 2020-02-23
Use linters by lint-staged

### LINKED ISSUE #579: Use js-beautify to format JS code
By using [js-beautify](https://www.npmjs.com/package/js-beautify), we can easily format our JS code. e.g.  `npm run format`

minggangw:
This patch changes:

- Use husky to hook pre-submit
- Use lint-staged
- Replace js-beautify with prettier, as it can handle both .js and .ts

Fix #579


.eslintrc
```diff
@@ -13,7 +13,7 @@ rules:
   comma-spacing: [2, {before: false, after: true}]
   comma-style: [2, "last"]
   handle-callback-err: [2, "^.*(e|E)rr" ]
-  indent: ["error", 2, {"SwitchCase": 1, "ArrayExpression": "off", "ObjectExpression": "off"}]
+  indent: ["error", 2, {"SwitchCase": 1}]
```
mattrichard, line 16: Instead of trying to match the lint rules with Prettier, consider using [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier). It will make sure your lint rules don't conflict with prettier. After installing it just add `"extends": ["prettier"]` to the *bottom* of the .eslintrc file.

mattrichard (commented):
I'm a big fan of prettier and use it in all my projects.

Just a heads up in case you haven't used prettier before. It's extremely opinionated and won't exactly do what you want (or prefer) maybe 10% of the time. But the remaining 90% definitely makes it worth it.

.eslintrc
minggangw: Thanks for pointing it out, that's reasonable. I'm going to update it.

.eslintrc
```diff
@@ -13,7 +13,7 @@ rules:
   comma-spacing: [2, {before: false, after: true}]
   comma-style: [2, "last"]
   handle-callback-err: [2, "^.*(e|E)rr" ]
-  indent: ["error", 2, {"SwitchCase": 1}]
+  indent: ["error", 2]
```
minggangw: Hi @mattrichard It seems that the `indent` is not overridden by the "prettier', would you please take a look if it's configured correctly? (currently, `npm run lint` has errors, pls see the CI output)

mattrichard: I was mistaken. Your rules will override eslint-config-prettier.

We should still extend "prettier", but then we'll need to adjust the rest of the rules to make eslint happy.

PR #586, opened 2020-02-26
Fix format script on Windows and include subfolders

mattrichard:
Files within sub-folders were skipped by the `format` script and Windows doesn't seem to like single quoted strings.

All changes in this PR aside from the package.json are from Prettier.


minggangw:
Thanks for making the improvement, LGTM.

PR #588, opened 2020-02-29
Add all action node bindings

mattrichard:
Adds the node binding methods required for action support.


src/rcl_bindings.cpp
```diff
@@ -1519,6 +2162,33 @@ BindingMethod binding_methods[] = {
     {"countPublishers", CountPublishers},
     {"countSubscribers", CountSubscribers},
     {"serviceServerIsAvailable", ServiceServerIsAvailable},
+    {"actionCreateClient", ActionCreateClient},
```
mattrichard: I was considering moving all action node bindings to separate file (e.g. rcl_action_bindings.cpp) as this file is massive already.

Any preference @minggangw?

src/rcl_bindings.cpp
```diff
@@ -1450,6 +1451,648 @@ NAN_METHOD(ServiceServerIsAvailable) {
...
+      rcl_action_get_zero_initialized_cancel_response();
+
+  rcl_ret_t ret = rcl_action_process_cancel_request(
+      action_server, cancel_request, &cancel_response);
+  if (ret != RCL_RET_OK) {
+    rcl_action_cancel_response_fini(&cancel_response);
+    Nan::ThrowError(GetErrorMessageAndClear());
+  }
+
+  *response = cancel_response.msg;
```
mattrichard: Out of all these changes, this is the one line I was unsure of the most. There's no existing example of how to handle this scenario. The `rcl_action_process_cancel_request` function above expects the response message to be wrapped in a `rcl_action_cancel_response_t` struct meaning I can't pass the response buffer (info[2]) in by reference, so I need to copy it afterwards.

Here's the rclpy usage of this function.

```python
# Cancel response is returned
cancel_response = _rclpy_action.rclpy_action_process_cancel_request(
    self._handle, cancel_request, self._action_type.Impl.CancelGoalService.Response)
```

Here's what I did for rclnodejs:

```javascript
// No return value, cancel response pointer is passed in then deserialized afterwards.
rclnodejs.actionProcessCancelRequest(this.handle, request.serialize(), response.toRawROS());
let cancelResponse = new Response();
cancelResponse.deserialize(response.refObject);
```

This works, but I'm wondering if the `cancel_reponse` memory needs to be cleaned up. If I call `rcl_action_cancel_response_fini(&cancel_response)` after this line then canceling the action doesn't work anymore.

Any thoughts on what (if anything) needs to be done here?

src/rcl_bindings.cpp
minggangw: Yes, this file expands too large, please split it into another `.cpp`.

src/rcl_bindings.cpp
minggangw: How about doing this, in cpp:
```cpp
// Allocate struct rcl_action_cancel_response_t and will be destoried in RclHandle::Reset()
rcl_action_cancel_response_t* cancel_response_ptr =  reinterpret_cast<rcl_action_cancel_response_t *>(malloc(sizeof(rcl_action_cancel_response_t )));
*cancel_response_ptr  = cancel_response;
auto js_obj = RclHandle::NewInstance(cancel_response_ptr , nullptr, [cancel_response_ptr] {
    rcl_action_cancel_response_fini(cancel_response_ptr );
});
info.GetReturnValue().Set(js_obj);
```

and in JavaScript

```js
let msg = rclnodejs.actionProcessCancelRequest(this.handle, request.serialize(), response.toRawROS());
let cancelResponse = new Response();
cancelResponse.deserialize(response.refObject);
msg.release();  // this triggers RclHandle::Reset() and will call rcl_action_cancel_response_fini()
```

I haven't tested this, just a potential method.

mattrichard: Thanks. I'll play around with it and let you know when everything is ready.

mattrichard: This helped me figure out what to do. Thanks. Will probably close the PR in favor of a different one with e2e action support.

PR #591, opened 2020-03-01
Add getLogger to Node object

mattrichard:
Adds getLogger function to the Node class. Both rclcpp and rclpy provide a get_logger function on the Node class, but it was missing in rclnodejs.

Actions make use of this, but it's a really a separate feature.

Fix #None


minggangw (approved):
Thanks for doing this to expose the log interface, LGTM

PR #592, opened 2020-03-01
Add getClock to Node object

mattrichard:
Adds an internal clock to rclnodejs nodes. In addition this also makes the following changes:

- Timer handles will now be children of the clock handle. This matches rclpy's implementation.
- Parameter events will use the node's clock instead of creating a new one each time.
- Exposes `ClockType` publicly.

Actions will also require use of the node's clock.

Fix #None

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -32,6 +32,7 @@ const Publisher = require('./publisher.js');
 const QoS = require('./qos.js');
 const Service = require('./service.js');
 const Subscription = require('./subscription.js');
+const TimeSource = require('./time_source.js');
 const Timer = require('./timer.js');
 
 const debug = require('debug')('rclnodejs:node');
@@ -90,6 +91,13 @@ class Node {
         this._parameterDescriptors.set(parameter.name, descriptor);
       }
     }
+
+    // Clock that has support for ROS time.
+    // Note: parameter overrides and parameter event publisher need to be ready at this point
+    // to be able to declare 'use_sim_time' if it was not declared yet.
+    this._clock = new Clock.ROSClock();
+    this._timeSource = new TimeSource(this);
+    this._timeSource.attachClock(this._clock);
 
     if (options.startParameterServices) {
       this._parameterService = new ParameterService(this);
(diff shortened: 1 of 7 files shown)


lib/node.js
```diff
@@ -458,6 +478,7 @@ class Node {
     }
 
     this.handle.release();
+    this._clock.handle.release();
```
mattrichard: The clock's handle should preferably be a child of the node's handle to avoid needed to add this here. There's two approaches we could take.

1. Allow an optional parent handle in the Clock's constructor, then update `rclnodejs.createClock` to take in this handle.
2. Expose `set_parent` on RclHandle as a NAN_METHOD to allow updating a handle's parent from JavaScript.

No. 2 is closer to rclpy's implementation, but I'm leaning more towards No. 1 as it keeps the handle management purely in c++.

lib/node.js
```diff
@@ -1036,7 +1065,7 @@ class Node {
       PARAMETER_EVENT_MSG_TYPE
     ))();
 
-    const secondsAndNanos = new Clock.ROSClock().now().secondsAndNanoseconds;
```
mattrichard, line 1039: qq. Since `new ROSClock()` calls malloc internally, wouldn't this be a memory leak since the clock has no parent handle and `release` is never called?

minggangw: If the `handle` doesn't have a parent, then the dtor will be called when this js object is GCed by v8, so the functional callback passed when creating this handle is called in `Reset`.
https://github.com/RobotWebTools/rclnodejs/blob/d3d5c2fe1ab4773121b6827076711ccebb159230/src/rcl_handle.cpp#L26-L28

lib/node.js
minggangw: If you want to call `rcl_clock_fini()` immediately after this line, we can accept option 1. 

Unlike other entities, e.g. pub/sub/client/service, they must stop working and release the resources before the `node` is destroyed, so their parent has to be a concrete node.  The Clock doesn't depend on a `node`, maybe we can let the GC take care of the timing of doing this, so just `this._clock = null`, how do you think?

lib/node.js
mattrichard: oh cool. I didn't know that. thanks

lib/node.js
mattrichard: I like the idea of setting `this._clock` to `null`. That's much simpler.

minggangw (approved):
Thanks for implementing this function LGTM

Issue #564, opened 2020-02-16
Rate support


mattrichard:
Support for creating a fixed rate is also missing in rclnodejs. This looks like it should be very simple to implement.

https://github.com/ros2/rclpy/pull/443

wayneparrott:
A quick thought:

This could be a design challenge for a single threaded javascript implementation based on the [rclpy Rate implementation](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/timer.py). The classic use-cases for Rate that I'm aware of is an endless loop that uses a Rate to establish a regular frequency for the loop. Following a Rate#sleep() call in this endless loop is a call to spin_once() that fires all active callbacks. **At issue is a deadlock that arises when using a Timer in the Rate implementation along with a spin_once() to trigger the Timer's callback**.

During Rate#sleep() the rate's timer.callback triggers an event which unblocks clients waiting in Rate#sleep() call. Unfortunately a **deadlock** situation arises as Timer.callback() will not be invoked unless the containing node is already spinning. Thus if the scenario is using spin_once() as in the code below, the Rate's timer is blocked forever waiting for its callback to be invoked by the spin_once(). But spin_once() will never be called.
```
import rclpy
from rclpy.executors import SingleThreadedExecutor
context = rclpy.context.Context()
rclpy.init(context=context)
node = rclpy.create_node('test_timer', context=context)
executor = SingleThreadedExecutor(context=context)
executor.add_node(node)
rate = node.create_rate(10)
while rclpy.ok(context=context):
  rate.sleep()
  executor.spin_once(10)
  print('rate test')
print ('done')
```

minggangw:
I read the comment for the `sleep()` method in rclpy
https://github.com/ros2/rclpy/blob/2b041106ea15d3e2824b00790f7b4b73eb699b4c/rclpy/rclpy/timer.py#L137-L143
It seems that **Rate** should be used carefully with`SingleThreadedExecutor`, and I looked through the test case which uses `SingleThreadedExecutor`
https://github.com/ros2/rclpy/blob/2b041106ea15d3e2824b00790f7b4b73eb699b4c/rclpy/test/test_rate.py#L28, it turns out the `sleep()` is called on a separate thread. Not sure how the usage of **Rate** with [MultiThreadedExecutor](https://github.com/ros2/rclpy/blob/2b041106ea15d3e2824b00790f7b4b73eb699b4c/rclpy/rclpy/executors.py#L687) looks like.

[...]

PR #593, opened 2020-03-06
Adds Rate class for #564

wayneparrott:
Implements a Rate timer based on the rclpy Rate timer. See issue #564.
Includes:
* update node, added node#createRate(hz)
* lib/rate.js
* types/rate.d.ts
* test/test-rate.js
* updated test/types/main.ts

Note for reviewers:
* See RateTimer in rate.js. This class runs a Timer in private rcl context.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -30,6 +30,7 @@ const {
 const ParameterService = require('./parameter_service.js');
 const Publisher = require('./publisher.js');
 const QoS = require('./qos.js');
+const Rates = require('./rate.js');
 const Service = require('./service.js');
 const Subscription = require('./subscription.js');
 const TimeSource = require('./time_source.js');
@@ -55,6 +56,7 @@ class Node {
     this._services = [];
     this._timers = [];
     this._guards = [];
+    this._rateTimerServer = null;
     this._parameterDescriptors = new Map();
     this._parameters = new Map();
     this._parameterService = null;
@@ -264,6 +266,36 @@ class Node {
     this.syncHandles();
 
     return timer;
+  }
+
+  /**
+   * Create a Rate.
+   *
+   * @param {number} hz - The frequency of the rate timer; default is 1 hz.
+   * @returns {Rate} - New instance
+   */
+  createRate(hz = 1) {
+    if (typeof hz !== 'number') {
+      throw new TypeError('Invalid argument');
+    }
+
+    const MAX_RATE_HZ_IN_MILLISECOND = 1000.0;
+    if (hz <= 0.0 || hz > MAX_RATE_HZ_IN_MILLISECOND) {
+      throw new RangeError(
+        `Hz must be between 0.0 and ${MAX_RATE_HZ_IN_MILLISECOND}`
+      );
+    }
+
+    // lazy initialize rateTimerServer
+    if (!this._rateTimerServer) {
+      this._rateTimerServer = new Rates.RateTimerServer(this);
+    }
+
+    const period = Math.round(1000 / hz);
+    const timer = this._rateTimerServer.createTimer(period);
+    const rate = new Rates.Rate(hz, timer);
+
+    return rate;
   }
 
   /**
@@ -485,6 +517,11 @@ class Node {
     this._clients = [];
     this._services = [];
     this._guards = [];
+
+    if (this._rateTimerServer) {
+      this._rateTimerServer.shutdown();
+      this._rateTimerServer = null;
+    }
   }
 
   /**
--- a/lib/rate.js
+++ b/lib/rate.js
@@ -1,0 +1,186 @@
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('bindings')('rclnodejs');
+const Context = require('./context.js');
+const NodeOptions = require('./node_options.js');
+
+const NOP_FN = () => {};
+
+/**
+ * A timer that runs at a regular frequency (hz).
+ *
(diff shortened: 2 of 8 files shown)


minggangw (commented):
Thanks for implementing Rate function, I didn't realize that we could instantiate another `ShadowNode` to create additional sub-thread, good idea. But I guess we still cannot block the Node.js main thread, see the comment. Shall we give a detailed explanation about what will happen when calling `sleep()`?

test/test-rate.js
```diff
@@ -0,0 +1,103 @@
...
+    rate.cancel(); // redundant cancel is a nop
+  });
+
+  it('rate sleep accuracy test, 1000 hz for 3 seconds', async function() {
+    // run 3 * hz,
+    //   collect and average the sleep intervals
+    //   compare average sleep interval with the period of the timer
+    const hz = 1000;
+    const rate = node.createRate(hz);
+
```
minggangw, line 92: I am wondering what it's the outcome if I insert:
```js
    setInterval(() => {
      // publish a top here.
    }, 10);
```

wayneparrott: Here's a simple experiment to see the effect of rate.
Start publishing msgs every 1 sec from commandline using:
```
ros2 topic pub topic 'std_msgs/msg/String' '{data: hello}'
```
Run this example to received msgs every 2 seconds:
```javascript
'use strict';

const rclnodejs = require('./index.js');

async function main() {
  await rclnodejs.init();
  const node = rclnodejs.createNode('test_node');
  const subscriptions = node.createSubscription(
    'std_msgs/msg/String',
    'topic', 
    undefined,
    msg => console.log(`Received(${Date.now()}): ${msg.data}`)
  );
  const rate = node.createRate(0.5);
  let forever = true;
  while (forever) { // should run every 2 secs
    await rate.sleep();
    rclnodejs.spinOnce(node);
  }
}

main();
```

minggangw: Yes, the subscription can work as expected, but publishing a topic can still be called through a callback of the `setInterval()`

wayneparrott: >but publishing a topic can still be called through a callback of the setInterval()

Correct. Is that not what you were expecting? Are you expecting Rate.sleep() to govern, i.e., block, calls to publish()?

I coded up the example with the setInterval( ()=>publisher.publish('hello'),10). When the interval invokes the callback, the published msg is available to all subscribers who's node is spinning or spinOnce. In this example  the rate loop is  running at 1/2 hz, while a msg is published every 10ms by the setInterval. With that, every 200th msg will be received when spinOnce is called. I believe the 199 msgs received during rate.sleep() are lost.

lib/rate.js
```diff
@@ -0,0 +1,204 @@
...
+  isCanceled() {
+    return this._rateTimer.isCanceled();
+  }
+}
+
+const NOP_FN = () => {};
+
+let NODE_CNT = 1;
+function genNodeName() {
+  return 'rate_context' + NODE_CNT++;
```
mattrichard: Should the name be prefixed with the name of the node that created the Rate? Otherwise, it looks like if you ran 2 separate JS scripts that create a Rate the names would conflict since NODE_CNT wouldn't be shared.

test/test-rate.js
minggangw: >>but publishing a topic can still be called through a callback of the setInterval()

>Correct. Is that not what you were expecting? Are you expecting Rate.sleep() to govern, i.e., block, calls to publish()?

The behavior (not block the loop) here is reasonable in Node.js, I just want to ensure when developers are calling `sleep()`, they are clear about the expectation (not a real sleep), thanks!

lib/rate.js
wayneparrott: Yes, absolutely! Good suggestion.
And/or we could make the Node hidden altogether.

wayneparrott:
PROPOSAL - refactor the Rate design as follows:
1. rename RateTimer to RateTimerServer, to produce Timers for use by Rate instances 
2. move RateTimerServer management from each rate instance into Node
3. use lazy init RateTimerServer on first Node#createRate() call
4. revamp Rate back to my 1st version where it is initialized with a Timer from the RateTimerServer during Node#createRate()

Tradeoffs of this refactor include:
1. Only 1 rcl context and node to support for all Rates created by a node - fewer moving parts, increases efficiency
2. Any RateTimerServer node initialization, etc is hidden from developers working with rate - simplifies use 
3. Increases Node's responsibility and internal complexity

Apologies for pushing this proposal after you've invested time reviewing the current impl. If approved I prefer making this change immediately before merging. Doing it in the future will result in a breaking api change for Rate.

Thoughts?

minggangw:
The redesign sounds more reasonable, I like using only one node to do it. We could also land this PR in advance and create a new issue to target the refactoring, as the changes are on `develop` I think it's fine, it's up to you. Please go ahead, thanks a lot!

wayneparrott:
The latest commit #b0fedaa is the refactored version of Rate with Node now managing a lazy initialized RateTimerServer, 1 per node instance. The RateTimerServer class is located in node.js. 

Per @mattrichard suggestion the RateTimerServer manages a private rcl context and node. This node shares the name of its container node as `_$(containing_node.name())_rate_timer_server`. The '_' prefix will keep it hidden from commandline operations such as `ros2 node list`.

Apologies for the late refactor after you've already invested review time. I'm much more positive about this version that the former.

minggangw (approved):
Thanks for refactoring the Rate implementation quickly, LGTM @mattrichard any comments? If everything is OK, I'm going to merge this PR, thanks!

lib/node.js
```diff
@@ -1208,4 +1245,70 @@ class Node {
   }
 }
 
+/**
+ * Internal class that creates Timer instances in a common private rcl context
+ * for use with Rate. The private rcl context ensures that Rate timers do not
+ * deadlock waiting for spinOnce/spin on the main rcl context.
+ */
+class RateTimerServer {
```
mattrichard: Since this only relates to Rates, can this class go in the `rate.js` file instead? The `node.js` file is already very large.

Other than that it LGTM.

wayneparrott: Agree node.js is getting large and I originally I implemented RateTimeServer in rate.js for that reason. But at the last min I moved it to node.js to keep it private and avoid exporting it from Rate. I'm fine with moving RateTimeServer to rate and exporting it.

wayneparrott: The latest commit includes RateTimeServer relocated from node.js to rate.js

mattrichard (approved):
LGTM

PR #596, opened 2020-03-07
Add ActionClient and ActionServer

mattrichard:
Adds the ActionClient and ActionServer providing full e2e support for actions. The implementation is almost identical to rclpy, only using a more JS approach like the rest of rclnodejs. The only big difference is that Promises are used over a custom "Future" class. This makes the action interface much more friendly and allows for using async/await.

See the examples included for sample usage.

Remaining work after this PR:
- Generate the TypeScript typings for action messages.
- Add action client & server TypeScript typings.
- Add action graph query functions.

Note this PR targets the "actions" branch.


minggangw:
The implementation looks good to me, thanks. One question, have you tested it with rclpy as an action client or action server?

mattrichard:
> One question, have you tested it with rclpy as an action client or action server?

Yes. Would you like cross-language testing as part of the actions unit testing? Behind the scenes actions just use plain messages and services, which I believe already has cross-language testing.

minggangw:
Not really, because you may notice that some cross-language tests lead to time out on the CIs sometimes, just want to ensure we have tested with other clients at least, thanks!

PR #597, opened 2020-03-07
Allow installing with Node v12 and update README

mattrichard:
Update package.json to allow installing rclnodejs with Node.js v12, and reflect allowed versions in the README.

Fix #None


minggangw (approved):
Thanks for the PR, LGTM!

PR #600, opened 2020-03-09
fix possible error when generating typescript definitions

koonpeng:
small change to fix error when generating ros-gazebo messages.

ros-gazebo includes services that end with `Request` like `BodyRequest` and `LinkRequest`. This causes the check to determine if an idl is a service to fail as it only checks if it ends with `Request`.


minggangw (approved):
Thanks for submitting the PR, merging.

Issue #598, opened 2020-03-08
Access current time from Node


wayneparrott:
In ROS2 time keeping is the responsibility of Node. We recently added getClock() to Node. In addition to accessing the clock is convenient access of the current time. This issue proposes adding `now(): Time` method to Node similar to the now() method of the [rclcpp.Node](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node.cpp) class.

PR #605, opened 2020-03-11
Added Node#now() with test #598

wayneparrott:
Adds convenient method to Node for accessing the current time.
- added now() method to Node that returns current Time using the node's clock
- added now() to node.d.ts
- added test case to test-node.js


minggangw (approved):
Thanks for the PR, merging...

PR #616, opened 2020-03-21
Add action TypeScript typings

mattrichard:
This PR contains the following changes.

- Adds typings for ActionClient and ActionServer.
- Generates action interface types.
- Replaces use of 'type' with 'interface' where applicable, which is generally preferred.
- Rename "Wrapper" to "Constructor" to follow TS naming conventions.

There are many more enhancements that can be made, but they are not necessary at this time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "scripts": {
     "install": "node-gyp rebuild",
     "docs": "cd docs && make",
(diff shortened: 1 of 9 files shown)


mattrichard:
The Travis CI build failure looks unrelated to these changes.

minggangw:
I see and we could ignore it or re-run it when the binary package of ROS2 is updated.

minggangw (approved):
LGTM, thanks!

PR #618, opened 2020-03-21
Fix all build warnings when building with Node.js v12

mattrichard:
Removes deprecated use of Set and Get.

Fix #None


minggangw (approved):
Thanks for fixing these warnings, LGTM.

PR #622, opened 2020-03-23
Fix test failures with dtslint v3.4

mattrichard:
All options removed are automatically implied by `strict: true`.

Fix #None


minggangw (approved):
LGTM, thanks for the PR and merging.

PR #623, opened 2020-03-24
Disable common compiler warnings

wayneparrott:
This PR continues the work started by #618 to remove compiler warnings.
- Removes warnings for comparing unsigned and signed ints
- Removes warnings for unused function return value


minggangw (approved):
Finally, we eliminate all warnings! Thanks for your effort. LGTM!

PR #627, opened 2020-03-25
Removes compiler format-string-literal warning

wayneparrott:
Adds a literal "%s" format string which satisfies the compiler and eliminates the following compiler warning: 
```
../src/rcl_bindings.cpp: In function ‘Nan::NAN_METHOD_RETURN_TYPE rclnodejs::Log(Nan::NAN_METHOD_ARGS_TYPE)’:
../src/rcl_bindings.cpp:1390:75: warning: format not a string literal and no format arguments [-Wformat-security]
     rcutils_log(&logging_location, severity, name.c_str(), message.c_str());
```


minggangw (approved):
Thanks for fixing it, LGTM

PR #645, opened 2020-04-06
Remove invalid import

wayneparrott:
Fixes Windows CI build error.

### DIFF
--- a/test/test-rate.js
+++ b/test/test-rate.js
@@ -78,6 +78,7 @@ describe('rclnodejs rate test suite', function() {
 
   it('rate server node naming test', async function() {
     const rate = node.createRate();
+    await assertUtils.createDelay(1000);
     const rateTimerServerName = `_${node.name()}_rate_timer_server`;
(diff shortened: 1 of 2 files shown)


wayneparrott:
Thoughts on the the frequent CI test failure to detect the hidden node used by Rate in test-rate.js test suite? We could add a bit of delay between Rate creation and request for nodes if you think that will improve reliability of this test. 
```
1) rclnodejs rate test suite
       rate server node naming test:
```

minggangw:
Of course, we should add it to reduce the failures. Another option, we can add the flaky test into https://github.com/RobotWebTools/rclnodejs/blob/develop/test/blacklist.json based on a specific platform to disable a test. But I think adding a delay is enough.

wayneparrott:
Fingers crossed the 1 sec delay improves the reliability of the rate test suite across all CI envs.

minggangw (approved):
Thanks for fixing this, merging.

Issue #658, opened 2020-05-28
generated msg declarations in typescript do not include typed-arrays


wayneparrott:
This is a general question about how typescript declarations for ros msgs represent numeric array properties. For example the `data` property of a ros2 image is defined as  uint8[] in [IDL](https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/msg/Image.msg). Yet the typescript interface generated by rclnodejs (see below) declares the data property as a `number[]`. 
```
export interface Image {
        header: std_msgs.msg.Header;
        height: number;
        width: number;
        encoding: string;
        is_bigendian: number;
        step: number;
        data: number[];
 }
```
Shouldn't `data` in the interface be defined as `number[] | Uint8Array` ?

minggangw:
> Shouldn't data in the interface be defined as number[] | Uint8Array ?

I agree as rclnodejs can adopt both types.

wayneparrott:
@minggangw @mattrichard Please review.

Proposed mapping of [IDL numeric type arrays](https://design.ros2.org/articles/idl_interface_definition.html) to JavaScript typed-array:

| IDL type | JS |
|---------|----|
| int8[]     | Int8Array|
| uint8[]     | Uint8Array|
| int16[]     | Int16Array|
| uint16[]     | Uint16Array|
| int32[]     | Int32Array|
| uint32[]     | Uint32Array|
| int64[] | BigInt64Array |
| uint64[] | BigUint64Array |
| float[] | Float32Array |
| double[] | Float64Array |
| long double[] | ? |  

The plan is to generate optional TS type decls for numeric arrays. 
Example:
```
data: number[] | Uint8Array;
```

wayneparrott:
update: hoping to submit a PR next week (june 24-ish)

wayneparrott:
fixed.

PR #665, opened 2020-06-27
Gen 'number[] | typed-array' union in tsd files #658

wayneparrott:
rostsd_gen/index.js
* Previously numeric idl array types were generated as a simple `number[]`. Now some of them are generate as a type union of `number[] | <typed-array>`

Implemented mapping of [IDL numeric type arrays](https://design.ros2.org/articles/idl_interface_definition.html) to JavaScript typed-array. Additional support was found by inspecting the the [idl parser code](https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/definition.py).

| IDL type | JS type generated |
|---------|--------------------|
| byte[]       | number[] \| Uint8Array |
| octet[]     |number[] \| Uint8Array |
| uint8[]  | number[] \| Uint8Array |
| int8[]     | number[] \| Int8Array|
| int16[]     | number[] \| Int16Array|
| short[]     | number[] \| Int16Array|
| uint16[]     | number[] \| Uint16Array|
| unsigned short| number[] \| Uint16Array|
| int32[]     | number[] \| Int32Array|
| long[]     | number[] \| Int32Array|
| uint32[]     | number[] \| Uint32Array|
| unsigned long[]     | number[] \| Uint32Array|
| float[] | number[] \| Float32Array |
| float32[] | number[] \| Float32Array |
| double[] | number[] \| Float64Array |
| float64[] | number[] \| Float64Array |
| int64[] | number[] |
| uint64[] | number[] |
| long long | number[] |
|unsigned long long| number[] |

test/types/main.ts
* added tests to support new number array union types, e.g., number[] |
typed-array

### DIFF
--- a/rostsd_gen/index.js
+++ b/rostsd_gen/index.js
@@ -406,7 +406,7 @@ function primitiveType2JSName(type) {
     case 'char':
     case 'byte':
     case 'octet':
-    
+
     // signed explicit integer types
     case 'short':
     case 'long':
@@ -461,6 +461,7 @@ function fieldTypeArray2JSTypedArrayName(type) {
     case 'uint8':
       jsName = 'Uint8Array';
       break;
+    case 'char':
     case 'int8':
       jsName = 'Int8Array';
       break;
--- a/rostsd_gen/index.js
+++ b/rostsd_gen/index.js
@@ -312,11 +312,23 @@ function saveMsgFieldsAsTSD(
       fieldType = 'any';
       tp = '';
     }
+    
     const tmpl = indentString(`${field.name}: ${tp}${fieldType}`, indent);
     fs.writeSync(fd, tmpl);
     if (field.type.isArray) {
       fs.writeSync(fd, '[]');
+     
+      if (fieldType === 'number') {
+        // for number[] include alternate typed-array types, e.g., number[] | uint8[]
+        let jsTypedArrayName = 
+          fieldTypeArray2JSTypedArrayName(field.type.type);
+
+        if (jsTypedArrayName) {
+          fs.writeSync(fd, ` | ${jsTypedArrayName}`);
+        }
+      }
     }
+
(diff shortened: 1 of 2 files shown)


minggangw:
Thanks for making this improvement! I found that the basic types provided by [IDL specification](https://design.ros2.org/articles/idl_interface_definition.html) have a wider range of types than the [ones](https://github.com/ros2/common_interfaces/tree/master/std_msgs#primitive-types) defined by ROS2 practically.

But, currently, we only consider the types defined by `std_msgs` in `rosidl_gen`, see below:

https://github.com/RobotWebTools/rclnodejs/blob/f19dc6a8be3eb5f741241dc4370769a087ef1cc2/rosidl_gen/templates/message.dot#L70-L95

So I am considering if we are going to define them consistently, as we will call the JS functions eventually. I prefer we adapt the IDL spec to extend the types defined in  `rosidl_gen`, what do you think?

wayneparrott:
> I prefer we adapt the IDL spec to extend the types defined in rosidl_gen, what do you think?

+1

I also found it helpful to review the code for ros2 idl compiler for the ground-truth for all the idl types recognized. 

Thanks for share the idl type -> typed-array mapping snippet. I notice that my PR does not address idl `char` array types. I'll post a revision with that change.

Also I wanted to share that [std_msgs package is deprecated as of Foxy](https://github.com/ros2/common_interfaces/pull/116). std_msgs is still included in the foxy distro with a deprecation notice.  The alternative example_interfaces package is now shipped in foxy.

minggangw:
> Also I wanted to share that std_msgs package is deprecated as of Foxy. std_msgs is still included in the foxy distro with a deprecation notice. The alternative example_interfaces package is now shipped in foxy.

Thanks for sharing this, I didn't notice that.

minggangw (approved):
Thanks for the patch and merging...

PR #666, opened 2020-06-29
fix compatibility for eloquent

koonpeng:
fix: compatibility for eloquent

v0.14.1 introduces support for ROS2 foxy however it also drops support for eloquent. This breaks existing users on v0.14.0+eloquent that expects api to remain stable when upgrading to a patch release.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -7,6 +7,9 @@
       },
     }
   },
+  'variables': {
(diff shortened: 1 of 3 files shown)


minggangw (approved):
Thanks for fix the compatibility problem! I reckon that will definitely meet the requirement from those who still want to work on eloquent, thanks!

Issue #668, opened 2020-06-30
Enable rclnodejs to fully support rcl contexts


wayneparrott:
This proposal is to refactor the rclnodejs design to properly support rcl contexts and to better use context in certain api's. My rationale is that the current rclnodejs design effectively restricts users to a single context. Yet at the rcl level, multiple contexts are supported. This is not a big deal in most cases but unnecessarily restricts more advance uses of rcl context such as the rclnodejs implementation of the Rate class which is hacky workaround. And the current visibility of context in a number of api can be refined to only a limited set of locations making for cleaner api.

**Benefit**
* cleaner user api
* conceptually maps to rcl closely
* can create and use multiple contexts, such as how we implemented Rate in it's own context

**Background**
A rclnodejs Context is a JS interface to the underlying rcl context. I like to think of the rcl context as a container for rcl objects such as nodes, publishers, actions, etc... The rcl design is such that it can host multiple contexts within a process which can lead to some clever advanced uses. The design of rclnodejs restricts us to a single instance of context. Specifically, the rclnodejs singleton maintains a global list of all nodes, separate from a node's context. When rclnodejs.shutdown(context) is called, all nodes regardless of their context (container) are stopped which is not typically desired.  Additionally context is passed as an argument to several node methods for which the node's implicit context reference should be used. 

**Recommendations:**
0. Overall goal is to limit the visibility and use of context to only the absolute essential places of dependency. And to do this with minimal changes and disruption such that all but most advanced uses of the current api are impacted.

1. Revise rclnodejs to properly manage the relationship between each context and nodes created in that context. 

3. Remove the context argument from `Node.createTimer() and `Node.createGuard()`  and use the node's implicit context reference (i.e., instance var).

4. Clean up the hacked Rate implementation to use it's own context without resorting to use of private api. 

**Concerns**
A main concern is breaking api changes. I suspect that removing the context parameters from api will not impact many users because most user of the api have relied on the default value of the context parameter. We could start with a deprecation warning where applicable that context parameter is not longer used and will be removed in a future update.

Thoughts?

minggangw:
Thanks for doing a deep investigation about the `Context` concept in rcl and rclnodejs and analyzing the steps to implement it. I did a simple implementation to support context, using the default `context` to pass to the `rcl` functions, so the `rclnodejs` cannot support multi-context. If we could achieve this feature, developers could customize their nodes by using a different context instance as you said.

[...]

PR #669, opened 2020-07-01
multiple context support #668

wayneparrott:
Adds the ability to support multiple simultaneous rcl contexts, #668.

index.js
* Replaced _nodes array with a Map<context,Array<Node>. A list of nodes
is maintained for each context provided to rcl.init(context,argv).
* Moved to end of file the import of node.js and time_source.js, the
TimeSource property and ShadowNode inheritance function call. This
change was made to workaround a circular dependency in rate.js.
* Added context parameter to
`isShutdown(context=Context.defaultContext())`

node.js
* Changed createRate() to return Promise<Rate>
* Removed the context parameter from createTimer()

rate.js
* Reimplemented RateTimerServer to use public rclnodejs methods.
* Added `init(): Promise<void>` method which creates internal private
context and node

test-init-shutdown.js
* Updated to include a multiple-contexts testcase

test-rate.js
* Updated to use node.createRate(): Promise<Rate> correctly.

node.d.ts
* Removed context parameter from createTimer() method declaration.

main.ts
* Updated Rate tests for async node.createRate()

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,6 @@ const fs = require('fs');
 const generator = require('./rosidl_gen/index.js');
 const loader = require('./lib/interface_loader.js');
 const logging = require('./lib/logging.js');
-// const Node = require('./lib/node.js');
 const NodeOptions = require('./lib/node_options.js');
 const {
   FloatingPointRange,
@@ -58,8 +57,6 @@ function inherits(target, source) {
   });
 }
 
-// inherits(rclnodejs.ShadowNode, Node);
-
 function getCurrentGeneratorVersion() {
   let jsonFilePath = path.join(generator.generatedRoot, 'generator.json');
 
@@ -90,11 +87,10 @@ function getCurrentGeneratorVersion() {
  */
 let rcl = {
   
-  // flag identifies when module has been init'ed
-  initialized: false,
-  
+  _rosVersionChecked: false,
+
   // Map<Context,Array<Node>
-  _nodes: new Map(),
+  _contextToNodeArrayMap: new Map(),
 
   /** {@link Clock} class */
   Clock: Clock,
@@ -195,7 +191,7 @@ let rcl = {
       throw new TypeError('Invalid argument.');
     }
 
-    if (!this._nodes.has(context)) {
+    if (!this._contextToNodeArrayMap.has(context)) {
       throw new Error('Invalid context. Must call rclnodejs(context) before using the context');
     }
 
@@ -210,7 +206,7 @@ let rcl = {
       namespace
     );
 
-    this._nodes.get(context).push(node);
+    this._contextToNodeArrayMap.get(context).push(node);
     return node;
   },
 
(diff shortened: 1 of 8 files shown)


wayneparrott:
Note how the appveyor ci build failed during the generate_messages.js script https://ci.appveyor.com/project/minggangw/rclnodejs/builds/33854406

I've seen similar error a few times (seg fault) when installing rclnodejs to a project with foxy debian package installation. I've not had time to investigate the cause yet.

minggangw:
Paste the error below:

```bash
Start JavaScript message generation...
Generation complete.
npm ERR! code ELIFECYCLE
npm ERR! errno 3221225477
npm ERR! rclnodejs@0.14.1 postinstall: `node scripts/generate_messages.js`
npm ERR! Exit status 3221225477
npm ERR! 
npm ERR! Failed at the rclnodejs@0.14.1 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\appveyor\AppData\Roaming\npm-cache\_logs\2020-07-01T19_05_42_428Z-debug.log
Command exited with code -1073741819
```

I've seen this before too, but it seems that the log doesn't give much meaningful information about the cause. I only found this error on Windows (appveyor), not ubuntu (build ROS2 from scratch).

P.S. `node scripts/generate_messages.js` generates JS messages from `.dot` template based on the IDL parser.

index.js
```diff
@@ -90,8 +89,12 @@ function getCurrentGeneratorVersion() {
...
  */
 let rcl = {
-  _initialized: false,
-  _nodes: [],
+  
+  // flag identifies when module has been init'ed
+  initialized: false,
+  
+  // Map<Context,Array<Node>
+  _nodes: new Map(),
```
minggangw: Could we rename |_nodes| more meaningful, like _contextToNodeArrayMap?

index.js
```diff
@@ -292,37 +303,47 @@ let rcl = {
...
+    if (this.isShutdown(context)) {
+      throw new Error('The module rclnodejs has been shutdown.');
+      return;
     }
 
-    this._nodes.forEach(node => {
+    // check for non-existant or deleted context
+    // if (context === Context.defaultContext()) {
+    //   console.log('default ctx', this._nodes.has(Context.defaultContext()));
+    // }
```
minggangw: Are these comment still necessary here?

index.js
```diff
@@ -227,6 +236,10 @@ let rcl = {
         throw new TypeError('argv elements must not be null');
       }
 
+      // setup internal state for context outside promise. 
+      rclnodejs.init(context.handle(), argv);
+      this._nodes.set(context, []);
+
       let that = this;
       if (!this._initialized) {
```
minggangw: If we are using the existence of a context, which is used to init this module, to judge if the module has initialized, I think we could remove |_initialized| flag now.

index.js
```diff
@@ -24,7 +24,7 @@ const fs = require('fs');
 const generator = require('./rosidl_gen/index.js');
 const loader = require('./lib/interface_loader.js');
 const logging = require('./lib/logging.js');
-const Node = require('./lib/node.js');
+// const Node = require('./lib/node.js');
```
minggangw: Please remove this comment.

index.js
```diff
@@ -59,7 +58,7 @@ function inherits(target, source) {
   });
 }
 
-inherits(rclnodejs.ShadowNode, Node);
+// inherits(rclnodejs.ShadowNode, Node);
```
minggangw: ditto

index.js
```diff
@@ -90,8 +89,12 @@ function getCurrentGeneratorVersion() {
  * @exports rclnodejs
  */
 let rcl = {
-  _initialized: false,
-  _nodes: [],
+  
+  // flag identifies when module has been init'ed
+  initialized: false,
```
minggangw: I don't find this property has been used, I guess you mean |_initialized|?

index.js
```diff
@@ -258,7 +269,7 @@ let rcl = {
             reject(e);
           });
       } else {
-        throw new Error('The module rclnodejs has been initialized.');
+        resolve();
```
minggangw, line 241: If we used the existence of context, we could remove this `else` I think

index.js
```diff
@@ -292,37 +303,47 @@ let rcl = {
...
+      throw new Error('The module rclnodejs has been shutdown.');
+      return;
     }
 
-    this._nodes.forEach(node => {
+    // check for non-existant or deleted context
+    // if (context === Context.defaultContext()) {
+    //   console.log('default ctx', this._nodes.has(Context.defaultContext()));
+    // }
+    if (!this._nodes.has(context)) {console.log('xxx'); return;}
```
minggangw: Could we have more descriptive wording here as log?

index.js
```diff
@@ -414,3 +435,13 @@ process.on('SIGINT', () => {
 });
 
 module.exports = rcl;
+
+// The following statements are located here to work around a 
+// circular dependency issue occuring in rate.js
```
minggangw, line 428: Is it possible to break the circular dependency? It seems tricky to do so.

minggangw (commented):
Thanks for implementing this feature, almost lgtm with some comments

index.js
wayneparrott: +1

index.js
wayneparrott: cruft. removed.

index.js
wayneparrott: removed cruft.

index.js
wayneparrott: opps! my bad. Yes it should be _initialized. Fixed

index.js
wayneparrott: I think a flag for indicating that the ros version has been checked (new idl files compiled when changed) has value. 
What do you think about renaming `_initialized` to `_rosVersionChecked` and using it in a slightly simplified init() function?

tl,dr; My initial thought was that we could replace `_initialized` with a  check for _contextToNodeArrayMap.size > 0 but this would not address an edge case where:
```
let context = new Context();
await rclnodejs.init(context);
rclnodejs.shutdown(context);
let newContext = new Context();
await rclnodejs.init(newContext); // a new version check would be conducted 
```
While very unlikely, I opted to retain and rename `_initialized` to `_rosVersionChecked`. Push back if you disagree.

index.js
wayneparrott: see my updated version.

index.js
wayneparrott: The work-around for the circular dependencies is to revert back to the former implementation of rate.js. That version of rate.js used private functions on node and duplicated code from index.js to avoid importing index.js. 

I have no issue rolling back to that code to avoid the complicated import/require code placement as we have now. Say the word and I'll revert rate.js and we will no longer need the current circular dependency workaround in index.js.

minggangw: I see, so let's keep this work around.

index.js
```diff
@@ -218,48 +217,54 @@ let rcl = {
...
-          .catch(e => {
-            reject(e);
-          });
-      } else {
-        throw new Error('The module rclnodejs has been initialized.');
+      // initialize context
+      rclnodejs.init(context.handle(), argv);
+      this._contextToNodeArrayMap.set(context, []);
+
+      if (this._rosVersionChecked) {
```
minggangw, line 239: I didn't realize that we could init with more than one context, it's reasonable to add this `if` and resolve the Promise earlier. Thanks for adding the corresponding ut.

minggangw (approved):
LGTM, thanks!

PR #679, opened 2020-07-18
Fix double-free when using typed array

### LINKED ISSUE #673: Possible use after free in callbacks
I created a new branch `bug/callback-uaf` to help reproduce it. It adds a `test-promise-gc` test case that fails with what looks like the result a use-after-free.

From my experiments, the uaf occurs when we assign the argument of of a callback that contains a typed array to a variable from the outer scope

```js
      const result = await new Promise((res) =>
        node.createSubscription(typeClass, topic, (msg) => {
          timer.cancel();
          node.destroy();
          res(msg);
        }),
      );
      global.gc();
```

here, `msg` is assigned to `result` through the promise resolver, after which we manually trigger the gc. The typed array ref in `result` now points to invalid memory.

Some points to note:
  * Without the gc call, the test passes but it may already refer to invalid memory.
  * ~Running in valgrind doesn't seem to catch anything.~
    * running in with the manual gc call does report invalid reads
  * In electron, this causes the array to contain all `0`s, even without the manual gc call, perhaps electron has different gc behavior.

Some initial debugging:

In `node.js`:

```js
  _runWithMessageType(Type, callback) {
    let message = new Type();

    callback(message.toRawROS(), () => {
      let result = new Type();
      result.deserialize(message.refObject);

      return result;
    });

    Type.destoryRawROS(message);
  }
```

`destroyRawROS` is called after the callback is ran, this eventually call the native function `CreateArrayBufferCleaner`. It creates a new `RclHandle`, which implements some RAII functions that `free` the underlying address when it goes out of scope. This `RclHandle` is wrapped in a `v8::Local<v8::Object>` and under a `EscapeHandleScope`.

  * ~Is there possibility for a `RclHandle` to get copied, and end up with multiple instance of `RclHandle` holding the same pointer, resulting in double frees when they go out of scope?~
    * I deleted the copy and move ctors and assignment operators and it still compiles.
  * The return value of `CreateArrayBufferCleaner` is not assigned to anything, would it cause the underlying `RclHandle` to go out of scope immediately, freeing the address?
  * Where does the address that `RclHandle` is holding comes from? Is there any chance we are freeing addresses that we are not managing?

minggangw:
This patch enables to transfer the ownership of ArrayBuffer to v8
when deserializing the messages, which has type of TypedArray.
This will avoid double-free when the v8 executes gc.

Fix #673

### DIFF
--- a/rosidl_gen/templates/message.dot
+++ b/rosidl_gen/templates/message.dot
@@ -460,7 +460,7 @@ class {{=objectWrapper}} {
     if (refObject.{{=field.name}}.size != 0) {
       {{=getWrapperNameByType(field.type)}}.ArrayType.freeArray(refObject.{{=field.name}});
       if ({{=getWrapperNameByType(field.type)}}.ArrayType.useTypedArray) {
-        deallocator.delayFreeStructMember(refObject.{{=field.name}}, {{=getWrapperNameByType(field.type)}}.refObjectArrayType, 'data');
+        // Do nothing, the v8 will take the ownership of the ArrayBuffer used by the typed array.
       } else {
         deallocator.freeStructMember(refObject.{{=field.name}}, {{=getWrapperNameByType(field.type)}}.refObjectArrayType, 'data');
       }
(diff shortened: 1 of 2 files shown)


koonpeng (approved):
thanks! confirmed it fixed the uaf and its working in electron's main process now too.

PR #684, opened 2020-07-27
fix: buffer overflows

koonpeng:
fix buffer overflow in `rclnodejs::GetBindingMethodsCount`
fix stack use after scope in `NAN_METHOD(Init)`
fix leak in `ShadowNode`

add instructions and helper scripts to run tests with asan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ install
 log
 .vscode
 .project
+
+# test build artifacts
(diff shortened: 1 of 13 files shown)


minggangw (approved):
Thanks for finding and fixing this kinda ancient bug and writing a detailed doc of how to use ASAN, which will help us to detect memory errors in the future. LGTM!

PR #685, opened 2020-07-27
fix: electron

koonpeng:
Workaround ENAMETOOLONG error when running in electron renderer process.

With this PR in electron https://github.com/electron/electron/pull/24659, rclnodejs is able to run in electron's main process and renderer process with `allowRendererProcessReuse` set to `false`.

### DIFF
--- a/src/addon.cpp
+++ b/src/addon.cpp
@@ -45,7 +45,7 @@ void InitModule(v8::Local<v8::Object> exports) {
     auto prog_name = program_invocation_name;
     auto end = strstr(prog_name, " -");
     assert(end);
-    prog_name[end - prog_name] = NULL;
+    prog_name[end - prog_name] = 0;
   }
 
   v8::Local<v8::Context> context = exports->GetIsolate()->GetCurrentContext();
--- a/src/addon.cpp
+++ b/src/addon.cpp
@@ -22,7 +22,32 @@
 #include "rcutils/macros.h"
 #include "shadow_node.hpp"
 
+bool IsRunningInElectronRenderer() {
+  auto global = Nan::GetCurrentContext()->Global();
+  auto process =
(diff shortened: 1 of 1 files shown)


koonpeng:
some info about the problem.

electron/chromium changes `argv[0]` to provide more helpful description in `ps`, however ros also uses argv[0] to determine the filename for logs. This often causes ros to use a filename that is too long for the system, resulting in ENAMETOOLONG error.

This is a workaround for it that terminates argv[0] on the first " -" occurrence so ros logging can create usable filenames. @minggangw Do you think if we should workaround it in rclnodejs or try to fix it in rcl? rcl is using `program_invocation_name` which is a glibc extension, I don't know if this affects windows or mac as it uses different ways to get the program name.

minggangw:
> Do you think if we should workaround it in rclnodejs or try to fix it in rcl?

As the usage of Electron is specific to rclnodejs only, I think it's fine to solve it inside rclnodejs itself.

>  rcl is using program_invocation_name which is a glibc extension

So, how rcl deals with the log file's name on Windows/macOS? What happens if we are running Election with rclnodejs on Windows when logging, will still meet ENAMETOOLONG?

koonpeng:
https://github.com/chromium/chromium/blob/f18e79d901f56154f80eea1e2218544285e62623/services/service_manager/embedder/set_process_title.cc#L88 It says here that windows and mac doesn't do anything so I believe that this only affects linux.

minggangw:
I see, so I reckon we will not have this kind of problem on Windows/macOS and this PR looks good now :+1:

minggangw (approved):
Thanks for working on this issue, lgtm and thanks!

minggangw:
@koonpeng would you please change the status to “Ready for review”, so I can merge this PR.

koonpeng:
I just updated to status. Thanks!

PR #690, opened 2020-07-31
Add missing type declarations for action server/client destroy methods

### LINKED ISSUE #687: No TypeScript support for ActionClient destroy() method
If you look at the [API documentation](http://robotwebtools.org/rclnodejs/docs/0.15.0/lib_action_client.js.html), a destroy method is documented. In the [type declaration file](https://github.com/RobotWebTools/rclnodejs/blob/develop/types/action_client.d.ts) no such method is documented. 

I'm opening to submitting a PR for this issue.

mattrichard:
Action servers and clients are not created through a node and therefore don't have corresponding destroy methods on the Node class. Instead, to destroy an action server/client outside of destroying the Node itself it is expected to call `destroy` on the server/client directly. This adds the type declarations for the `destroy` function which were missing.

Fix #687

### DIFF
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -367,6 +367,9 @@ client.isServiceServerAvailable();
 // $ExpectType Promise<boolean>
 actionClient.waitForServer();
 
+// $ExpectType void
+actionClient.destroy();
+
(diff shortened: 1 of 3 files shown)


minggangw (approved):
Thanks for adding this, lgtm!

PR #692, opened 2020-08-04
Correct the description of sendGoal function

### LINKED ISSUE #689: See When Action Completes
I can't figure out how to implement a callback on an `ActionClient` when the action completes, just on updates and when the action is successfully received. As mentioned in line 190 [here](http://robotwebtools.org/rclnodejs/docs/0.15.0/lib_action_client.js.html):

```
  /**
   * Send a goal and asynchronously get the result.
   *
   * The result of the returned Promise is set to a ClientGoalHandle when receipt of the goal
   * is acknowledged by an action server.
   *
   * @param {object} goal - The goal request.
   * @param {function} feedbackCallback - Callback function for feedback associated with the goal.
   * @param {object} goalUuid - Universally unique identifier for the goal. If None, then a random UUID is generated.
   * @returns {Promise} - A Promise to a goal handle that resolves when the goal request has been accepted or rejected.
   */
```

However, on line 313, I see the function `_getResult`:

```
  /**
   * Request the result for an active goal asynchronously.
   * @ignore
   * @param {ClientGoalHandle} goalHandle - Handle to the goal to cancel.
   * @returns {Promise} - A Promise that resolves when the get result request has been processed.
   */
```

Is this something that should be implemented, or an accepted restriction?

minggangw:
Fix #689

### DIFF
--- a/lib/action/client.js
+++ b/lib/action/client.js
@@ -188,10 +188,10 @@ class ActionClient extends Entity {
   }
 
   /**
-   * Send a goal and asynchronously get the result.
+   * Send a goal and wait for the goal ACK asynchronously.
    *
-   * The result of the returned Promise is set to a ClientGoalHandle when receipt of the goal
-   * is acknowledged by an action server.
+   * Return a Promise object that is resolved with a ClientGoalHandle when receipt of the goal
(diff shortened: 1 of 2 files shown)


minggangw:
@mattrichard PTAL, thanks!

minggangw:
I'd like to merge this PR in advance, so we could release v0.15.2 to include the recent bugfix, thanks!

PR #697, opened 2020-08-22
Changed TimeSource subscription to use topic `/clock`

### LINKED ISSUE #696: Error in TimeSource subscription to /clock topic
Currently when ROSTime is active via parameter `use_sim_time=true`, TimeSource incorrectly subscribes to the relative topic `clock`. Expected TimeSource to subscribe to the topic `/clock` according to https://design.ros2.org/articles/clock_and_time.html. Note that the back-slash char is missing.

@minggangw currently test-time-source.js is sending Time msgs to the topic `clock`. Shouldn't the topic be `/clock`? 

The background on this is I was attempting to incrementally step time for a node that is using the `use_sim_time` parameter. I used the `ros2 topic pub` tool to publish sim time updates to the node and the time updates where not being applied. Here are 2 example commands that failed until I changed the TimeSource topic to subscribe to `/clock` instead of `clock`.

```
ros2 topic pub /clock builtin_interfaces/msg/Time "{nanosec: 0, sec: 800}" -1
```
and
```
ros2 topic pub clock builtin_interfaces/msg/Time "{nanosec: 0, sec: 800}" -1
```

wayneparrott:
Changed TimeSource subscription and test-time-source.js to use topic `/clock` in place of `clock`. 

I initiated a discussion on issue #696 describing the need for this small change when using sim-time. I am proactively submitting this PR prior to agreement for the need to use topic `/clock`.

Fix #696

### DIFF
--- a/lib/time_source.js
+++ b/lib/time_source.js
@@ -22,7 +22,7 @@ const { Parameter, ParameterType } = require('./parameter.js');
 const Time = require('./time.js');
 
 const USE_SIM_TIME_PARAM = 'use_sim_time';
-const CLOCK_TOPIC = 'clock';
+const CLOCK_TOPIC = '/clock';
 
 /**
  * @class - Class representing a TimeSource in ROS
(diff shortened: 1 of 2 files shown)


minggangw (approved):
Thanks for fixing this problem, LGTM!

PR #703, opened 2020-09-29
Support to publish serialized messages

### LINKED ISSUE #646: Support to publish/subscribe serialized messages
rclnodejs should support the capability of publishing/subscribing serialized messages, which offers by rcl library:

- [rcl_publish_serialized_message](https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/include/rcl/publisher.h#L359)
- [rcl_take_serialized_message](https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/include/rcl/subscription.h#L308)

minggangw:
This patch implements the feature that a publisher can publish a raw
message, and an example is added.

Fix #646

### DIFF
--- a/example/publisher-raw-message.js
+++ b/example/publisher-raw-message.js
@@ -1,0 +1,43 @@
+// Copyright (c) 2020 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


wayneparrott:
Thx for the enhancement.  I did a quick read thru of the code and it looks good to me. Qualifier: my C is a little rusty so will defer to other reviewers.

The Publisher#publish() method declaration in types/publisher.d.ts needs to be updated to support the new alternative Buffer parameter type, i.e., `publisher(message: Message | Buffer) : void`. Do you want to include this in this PR?

minggangw:
> Do you want to include this in this PR?

@wayneparrott Would you please help to submit a separate PR for the ts declaration? I will merge it once this PR lands.

I have tested the implementation with the rclpy client and will add a unit test when the subscription part is done, thanks!

PR #720, opened 2020-11-11
shuts down all contexts on sigint; remove g_sigint_gc

koonpeng:
This changes the SIGINT handler to only run when no other handlers are installed, it also shuts down all contexts instead of just the default context.

This also fix a bug where a context is not properly shutdown in the sigint handler, with this bug fixed there should no longer be a need to explicitly call `process.exit` in the handler, we will let node exit gracefully when all work in the event loop are finished.

Discussion:
I make the signal handler run only when no other handlers are installed, I did this because maybe an application would want to handle SIGINT differently and not shutdown rclnodejs. But doing this has the side effect that if the app installs a signal handler to clean up it's own resource and it expects rclnodejs to cleanup itself (because that is the behavior when no handlers is installed), rclnodejs's handler would not run.

So given this scenario, I wonder if it is better to
* Run the handler only when it is the only one installed, has the side effect mentioned above.
* Run the handler regardless, apps that wants it's own SIGINT behavior that does not clean things up would have to remove rclnodejs's handler.
* Do not install signal handlers. Users may be confused why ctrl+c does not work by default.

### DIFF
--- a/index.js
+++ b/index.js
@@ -184,7 +184,7 @@ let rcl = {
     nodeName,
     namespace = '',
     context = Context.defaultContext(),
-    options = NodeOptions.defaultOptions,
+    options = NodeOptions.defaultOptions
   ) {
     if (typeof nodeName !== 'string' || typeof namespace !== 'string') {
       throw new TypeError('Invalid argument.');
@@ -208,6 +208,7 @@ let rcl = {
 
   /**
    * Init the module.
+   *
    * @param {Context} context - The context, default is Context.defaultContext().
    * @param {string[]} argv - Process commandline arguments.
    * @return {Promise<undefined>} A Promise.
@@ -405,15 +406,26 @@ let rcl = {
   createMessageObject(type) {
     return this.createMessage(type).toPlainObject();
   },
+
+  /**
+   * Removes the default signal handler installed by rclnodejs. After calling this, rclnodejs
+   * will no longer clean itself up when a SIGINT is received, it is the application's
+   * responsibility to properly shut down all nodes and contexts.
+   *
+   * Application which wishes to implement its own signal handler logic should call this.
+   * @returns {undefined}
+   */
+  removeSignalHandlers() {
+    process.removeListener('SIGINT', _sigHandler);
+  },
 };
 
 const _sigHandler = () => {
+  // shuts down all live contexts. Applications which wishes to use their own signal handlers
+  // should call `rclnodejs.removeSignalHandlers`.
   debug('Catch ctrl+c event and will cleanup and terminate.');
-  // only run when no other signal handlers are installed
-  if (process.listenerCount('SIGINT') === 1 && process.listeners('SIGINT')[0] === _sigHandler) {
-    for (let ctx of rcl._contextToNodeArrayMap.keys()) {
-      rcl.shutdown(ctx);
-    }
+  for (let ctx of rcl._contextToNodeArrayMap.keys()) {
+    rcl.shutdown(ctx);
   }
 };
 process.on('SIGINT', _sigHandler);
--- a/index.js
+++ b/index.js
@@ -184,7 +184,7 @@ let rcl = {
     nodeName,
     namespace = '',
     context = Context.defaultContext(),
-    options = NodeOptions.defaultOptions
+    options = NodeOptions.defaultOptions,
   ) {
     if (typeof nodeName !== 'string' || typeof namespace !== 'string') {
       throw new TypeError('Invalid argument.');
@@ -407,11 +407,16 @@ let rcl = {
   },
 };
 
-process.on('SIGINT', () => {
+const _sigHandler = () => {
   debug('Catch ctrl+c event and will cleanup and terminate.');
-  rcl.shutdown();
-  process.exit(0);
-});
+  // only run when no other signal handlers are installed
+  if (process.listenerCount('SIGINT') === 1 && process.listeners('SIGINT')[0] === _sigHandler) {
+    for (let ctx of rcl._contextToNodeArrayMap.keys()) {
+      rcl.shutdown(ctx);
+    }
+  }
+};
+process.on('SIGINT', _sigHandler);
 
 module.exports = rcl;
 
--- a/index.js
+++ b/index.js
@@ -419,12 +419,13 @@ let rcl = {
    * @returns {undefined}
    */
   removeSignalHandlers() {
+    // this will not throw even if the handler is already removed
     process.removeListener('SIGINT', _sigHandler);
   },
 };
 
 const _sigHandler = () => {
-  // shuts down all live contexts. Applications which wishes to use their own signal handlers
+  // shuts down all live contexts. Applications that wishes to use their own signal handlers
   // should call `rclnodejs.removeSignalHandlers`.
   debug('Catch ctrl+c event and will cleanup and terminate.');
   for (let ctx of rcl._contextToNodeArrayMap.keys()) {
--- a/index.js
+++ b/index.js
@@ -173,12 +173,13 @@ let rcl = {
   getActionNamesAndTypes: getActionNamesAndTypes,
 
   /**
-   * Create a node.
+   * Create and initialize a node.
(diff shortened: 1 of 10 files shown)


felixdivo:
See also #721, it is very much related

index.js
```diff
@@ -407,12 +407,6 @@ let rcl = {
   },
 };
 
-process.on('SIGINT', () => {
-  debug('Catch ctrl+c event and will cleanup and terminate.');
-  rcl.shutdown();
-  process.exit(0);
```
minggangw: The initial purpose of installing a handler and handling SIGINT here is that I once found the program cannot be terminated by `ctrl+c` sometimes and this gives a chance to quit the process. I'm not sure if we may still meet the issue.

minggangw: Finally, I remember the reason why we need to shut down here (considering the code was written several years ago...).

At that time, there is no kind of Context concept, so `rcl.shutdown()` will stop the sub-thread and the main thread could quit as expected. But now user may create more than one Context, which means more sub-threads are running simultaneously, so simply executing `rcl.shutdown()` is not enough and we must ensure all contexts are cleaned up properly when terminating.
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/index.js#L305

minggangw:
Discussion about SIGINT https://github.com/RobotWebTools/rclnodejs/pull/721#issuecomment-726650101

felixdivo:
I think we should not ignore `package-lock.json`;). As you said, controversial. But it's AFAIK best practice, and [the manual](https://github.com/npm/cli/blob/release-6.14.7/docs/content/configuring-npm/package-lock-json.md) says:

> This file is intended to be committed into source repositories, and serves various purposes:
>
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
>
>- Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
>
>- To facilitate greater visibility of tree changes through readable source control diffs.
>
>- And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

index.js
```diff
@@ -407,11 +407,16 @@ let rcl = {
   },
 };
 
-process.on('SIGINT', () => {
+const _sigHandler = () => {
   debug('Catch ctrl+c event and will cleanup and terminate.');
-  rcl.shutdown();
-  process.exit(0);
-});
+  // only run when no other signal handlers are installed
```
felixdivo: I really like this change and it should replace #721 for these lines. Two things though: The debug call in line 411 should be moved down into the if-check, and maybe we should document how this works in the module documentation? E.g something along:

> Only if no other handlers for `SIGINT` events are registered, *rclnodejs* will shut down all live contexts in these events. This is useful to gracefully exit the process when `CTRL+C` is pressed. Once other signal handlers are registered using `process.on('SIGINT', some_handler)`, the handler of *rclnodejs* will not attempt to shut down the contexts.

felixdivo: Alternatively, we could supply an extra argument to specify `auto_cleanup=False` to `rclnodejs.init(conetxt, argv)` to make it explicit. This might cause less confusion for users.

koonpeng:
Thanks for the comments. After some thoughts, I changed to default behavior to always cleanup rclnodejs. Reason is because I figured most apps would fall into one of these

* Does not have any long running background tasks, expects ctrl+c to "just work" to gracefully exit the program.
* Has some background tasks, so it installs some signal handlers to clean up it's resources.
* Has some special behavior when it receive SIGINT, it may only want to perform cleanup when some conditions are met.

The first 2 is probably the most common and the new behavior should work for them. Particularly for case 2 because we are no longer forcing node to exit, it should work well with the app's signal handler. For case 3, I added a new function `rclnodejs.removeSignalHandlers` to allow the app to perform it's own cleanup logic.

koonpeng:
> I think we should not ignore package-lock.json;). As you said, controversial. But it's AFAIK best practice, and the manual says:

This is a complicated topic, the docs recommend committing package-lock.json but there are some arguments which suggests why we should ignore it. See https://dev.to/gajus/stop-using-package-lock-json-or-yarn-lock-3ddi. Since package-lock is not committed in this repo, I assumed that is what rclnodejs prefer. Perhaps @minggangw can help explain if we should/should not ignored it?

minggangw:
I haven't met any problem when running `npm install` without package-lock.json until now.The [npm doc](https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json) said this file should be committed into the repo, but I think if everything runs well without it, we could ignore it now and add it back if something goes wrong in the future.

felixdivo:
> See https://dev.to/gajus/stop-using-package-lock-json-or-yarn-lock-3ddi.

Interesting point. Thank's for sharing. I agree with you now. Delete it and wait what happens. Maybe add the link to that article to the `.gitignore` file or something like that.

The npm docs should better discuss that...

index.js
```diff
@@ -409,13 +409,31 @@ let rcl = {
...
+  /**
+   * Removes the default signal handler installed by rclnodejs. After calling this, rclnodejs
+   * will no longer clean itself up when a SIGINT is received, it is the application's
+   * responsibility to properly shut down all nodes and contexts.
+   *
+   * Application which wishes to implement its own signal handler logic should call this.
+   * @returns {undefined}
+   */
+  removeSignalHandlers() {
+    process.removeListener('SIGINT', _sigHandler);
```
felixdivo, line 423:
```suggestion
    // this will not throw even if the handler is already removed
    process.removeListener('SIGINT', _sigHandler);
```

I tested it with a generic `EventHandler`.

index.js
```diff
@@ -409,13 +409,31 @@ let rcl = {
...
+  },
 };
 
-process.on('SIGINT', () => {
-  debug('Caught SIGINT/ctrl+c event and will cleanup and terminate.');
-  rcl.shutdown();
-  process.exit(0);
-});
+const _sigHandler = () => {
+  // shuts down all live contexts. Applications which wishes to use their own signal handlers
```
felixdivo:
```suggestion
  // shuts down all live contexts. Applications that wish to use their own signal handlers
```

index.js
```diff
@@ -409,13 +409,31 @@ let rcl = {
...
+  // should call `rclnodejs.removeSignalHandlers`.
+  debug('Catch ctrl+c event and will cleanup and terminate.');
+  for (let ctx of rcl._contextToNodeArrayMap.keys()) {
+    if (!rcl.isShutdown(ctx)) {
+      rcl.shutdown(ctx);
+    }
```
felixdivo:
```suggestion
    rcl.shutdown(ctx);
```

This check is already done within `rcl.shutdown()` since #721.

koonpeng:
Windows CI is failing sporadically, I wonder if it's because it's using node v12.19.0 which has the segfault bug. Also the cpp build failed, not sure if it's related.

felixdivo (approved):

minggangw (approved):
LGTM! Thanks for making this improvement and I think the handling of sigint is perfect now :)

minggangw:
```bash
tep 17/21 : RUN rosdep install --from-paths $ROS2_WS/ros2-linux/share --ignore-src --rosdistro foxy -y --skip-keys "console_bridge fastcdr fastrtps osrf_testing_tools_cpp poco_vendor rmw_connext_cpp rosidl_typesupport_connext_c rosidl_typesupport_connext_cpp rti-connext-dds-5.3.1 tinyxml_vendor tinyxml2_vendor urdfdom urdfdom_headers"
 ---> Running in 00f69051d49d
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
rqt_bag: Cannot locate rosdep definition for [rosbag2_transport_py]
```
There is an error on travis-ci now, I tried to fix it by #727, but it seems not workable. So, I'd like merge this PR in advance and get back to recheck travis-ci with the next package of https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/, thanks!

PR #721, opened 2020-11-11
Improve the docs & shutdown behavior when the node process is terminated

### LINKED ISSUE #719: Suppress error message at shutdown
Every time I close my node.js server with rclnodejs running I get this error message: 

```
.../my-project/node_modules/rclnodejs/index.js:300
      throw new Error('The module rclnodejs has been shutdown.');
      ^

Error: The module rclnodejs has been shutdown.
    at Object.shutdown (.../my-project/node_modules/rclnodejs/index.js:300:13)
    at process.<anonymous> (.../my-project/node_modules/rclnodejs/index.js:412:7)
    at process.emit (events.js:314:20)
```

It originates from here: https://github.com/RobotWebTools/rclnodejs/blob/178c2c7068e9c1961f0db3328ab74a366803340e/index.js#L410-L414

I don't quite get why it is raised in the first place, but anyhow, wouldn't it be better if there was a `rclnodejs.tryShutdown()` method that would be called in that event instead? It would be much like the existing `Context.tryShutdown()`: Don't do anything if the (default) context is already shut down and else shut it down now.

Is such a change welcome? If yes, I can offer to implement it.

felixdivo:
This PR:
- First of all fixes #719.
- Improves the documentation in `rclnodejs/index.js`.

Some other stuff that I noticed and wanted to discuss. In [`rclnodejs/lib/context.js`](https://github.com/RobotWebTools/rclnodejs/blob/develop/lib/context.js):

- [x] 1. Should we remove `Context.shutdownDefaultContext()`? I'd argue that we want to have a `Context.tryShutdownDefaultContext()` too, but that would start to weirdly duplicate stuff. Instead, I'd suggest the following:
    - Delete `Context.shutdownDefaultContext()`, which would now become the barely longer `Context.defaultContext().shutdown()`.
    - By that, we get the "try" version for free, e.g. with `Context.defaultContext().tryShutdown()`.
    - Finally, we should add a check like `if (this === defaultContext) { defaultContext = undefined; }` to `Context.shutdown()`/`Context.tryShutdown()` anyways, as we would miss the cleanup if a user already explicitly chose to shut down the default context. This would, however, make `Context.shutdownDefaultContext()` even more redundant.
    - If we were to implement it like that, would we have a deprecation cycle/notice?
    - It would slightly simplify `_shutdown()` in `rclnodejs/index.js`.
   - ==> Yes, done.
- [x] 2. Shouldn't `defaultContext` be set to `null` instead of `undefined` to explicitly signal that it is absent and not just a typo somewhere? ==> Yes, done.
- [x] 3. I would expect the code `rclnodejs.shutdown()` to shut down *all* existing contexts instead of only the default one. Is that intentional? Is there a will to change that? ==> Okay, I'm convinced to not change things here. We follow rclpy's semantics.
- [x] 4. Shouldn't we call `NodeOptions.defaultOptions` in [`index.js`](https://github.com/RobotWebTools/rclnodejs/blob/develop/index.js#L187) instead of reference that method? I.e. append `()`. ==> Nevermind, didn't know the getter syntax.

EDIT:

- [x] I slightly changed the signature of the `Context` class to better align with that of rclpy. Is that OK? ==> Apparently
- [x] How should the actual `exit()` be done? ==> Will be improved in #720 
- [X] Once all else is done, how should we notify the users of the changes? ==> place it in the release notes

### DIFF
--- a/index.js
+++ b/index.js
@@ -311,7 +311,7 @@ let rcl = {
 
   /**
    * Shuts down the given context by shutting down and destroying all nodes contained within.
-   * Does nothing is the context si already shut down.
+   * Does nothing if the context is already shut down.
    * @param {Context} [context=Context.defaultContext()] - The context to be shutdown.
    * @return {undefined}
    */
--- a/index.js
+++ b/index.js
@@ -178,7 +178,8 @@ let rcl = {
    * @param {string} namespace - The namespace used in ROS, default is an empty string.
    * @param {Context} context - The context, default is Context.defaultContext().
    * @param {NodeOptions} options - The options to configure the new node behavior.
-   * @return {Node} The instance of Node.
+   * @return {Node} A new instance of the specified node.
+   * @throws {Error} If the given context is not registered.
    */
   createNode(
     nodeName,
@@ -211,6 +212,7 @@ let rcl = {
    * @param {Context} context - The context, default is Context.defaultContext().
    * @param {string[]} argv - Process commandline arguments.
    * @return {Promise<undefined>} A Promise.
+   * @throws {Error} If the given context has already been initialized.
    */
   init(context = Context.defaultContext(), argv = process.argv) {
     return new Promise((resolve, reject) => {
@@ -263,6 +265,7 @@ let rcl = {
    * Start to spin the node, which triggers the event loop to start to check the incoming events.
    * @param {Node} node - The node to be spun.
    * @param {number} [timeout=10] - ms to wait, block forever if negative, don't wait if 0, default is 10.
+   * @throws {Error} If the node is already spinning.
    * @return {undefined}
    */
   spin(node, timeout = 10) {
@@ -279,6 +282,7 @@ let rcl = {
    * Execute one item of work or wait until a timeout expires.
    * @param {Node} node - The node to be spun.
    * @param {number} [timeout=10] - ms to wait, block forever if negative, don't wait if 0, default is 10.
+   * @throws {Error} If the node is already spinning.
    * @return {undefined}
    */
   spinOnce(node, timeout = 10) {
@@ -292,15 +296,40 @@ let rcl = {
   },
 
   /**
-   * @param {Context} context - The context to be shutdown.
+   * Shuts down the given context by shutting down and destroying all nodes contained within.
+   * @param {Context} [context=Context.defaultContext()] - The context to be shutdown.
    * @return {undefined}
+   * @throws {Error} If the context is already shut down.
    */
   shutdown(context = Context.defaultContext()) {
     if (this.isShutdown(context)) {
       throw new Error('The module rclnodejs has been shutdown.');
-      return;
+    } else {
+      this._shutdown(context)
     }
+  },
 
+  /**
+   * Shuts down the given context by shutting down and destroying all nodes contained within.
+   * Does nothing is the context si already shut down.
+   * @param {Context} [context=Context.defaultContext()] - The context to be shutdown.
+   * @return {undefined}
+   */
+  tryShutdown(context = Context.defaultContext()) {
+    if (!this.isShutdown(context)) {
+      this._shutdown(context)
+    }
+  },
+
+  /**
+   * Shuts down the given context by shutting down and destroying all nodes contained within.
+   * Does not perform any checks.
+   *
+   * @param {Context} context - The context to be shutdown.
+   * @return {undefined}
+   * @private
+   */
+  _shutdown(context) {
     // shutdown and remove all nodes assigned to context
     this._contextToNodeArrayMap.get(context).forEach((node) => {
       node.stopSpinning();
@@ -317,8 +346,8 @@ let rcl = {
   },
 
   /**
-   * A predictate for testing if a context has been shutdown.
-   * @param {Context} [context=defaultContext] - The context to inspect
+   * A predicate for testing if a context has been shutdown.
+   * @param {Context} [context=Context.defaultContext()] - The context to inspect
    * @return {boolean} Return true if the module is shut down, otherwise return false.
    */
   isShutdown(context = Context.defaultContext()) {
@@ -335,13 +364,13 @@ let rcl = {
   },
 
   /**
-   * Search packgaes which locate under path $AMENT_PREFIX_PATH, regenerate all JavaScript structs files from the IDL of
+   * Search packages which locate under path $AMENT_PREFIX_PATH, regenerate all JavaScript structs files from the IDL of
    * messages(.msg) and services(.srv) and put these files under folder 'generated'. Any existing files under
    * this folder will be overwritten after the execution.
    * @return {Promise<undefined>} A Promise.
    */
   regenerateAll() {
-    // This will trigger to regererate all the JS structs used for messages and services,
+    // This will trigger to regenerate all the JS structs used for messages and services,
     // to overwrite the existing ones although they have been created.
     debug('Begin regeneration of JavaScript code from ROS IDL files.');
     return new Promise((resolve, reject) => {
@@ -359,7 +388,7 @@ let rcl = {
   },
 
   /**
-   * Judge if the topic/service is hidden, see http://design.ros2.org/articles/topic_and_service_names.html#hidden-topic-or-service-names
+   * Judge if the topic/service is hidden (see [the ROS2 design documentation]{@link http://design.ros2.org/articles/topic_and_service_names.html#hidden-topic-or-service-names}).
    * @param {string} name - Name of topic/service.
    * @return {boolean} - True if a given topic or service name is hidden, otherwise False.
    */
@@ -397,10 +426,10 @@ let rcl = {
   },
 
   /**
-   * Create a plain JavaScript by specified type identifier
+   * Create a plain JavaScript from the specified type identifier.
    * @param {string|Object} type -- the type identifier, acceptable formats could be 'std_msgs/std/String'
    *                                or {package: 'std_msgs', type: 'msg', name: 'String'}
-   * @return {Object|undefined} A plain JavaScript of that type
+   * @return {Object|undefined} A plain JavaScript of that type or undefined if the object could not be created
    */
   createMessageObject(type) {
     return this.createMessage(type).toPlainObject();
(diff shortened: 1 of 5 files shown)


felixdivo:
cc @minggangw

index.js
```diff
@@ -292,15 +296,40 @@ let rcl = {
...
 
+  /**
+   * Shuts down the given context by shutting down and destroying all nodes contained within.
+   * Does nothing if the context is already shut down.
+   * @param {Context} [context=Context.defaultContext()] - The context to be shutdown.
+   * @return {undefined}
+   */
+  tryShutdown(context = Context.defaultContext()) {
+    if (!this.isShutdown(context)) {
+      this._shutdown(context);
```
koonpeng: This could still throw if the underlying rcl call fails. Also related to other `@throws` docs, they may throw for other reasons not specified in the doc.

index.js
```diff
@@ -409,14 +438,14 @@ let rcl = {
 
 process.on('SIGINT', () => {
   debug('Catch ctrl+c event and will cleanup and terminate.');
-  rcl.shutdown();
+  rcl.tryShutdown();
   process.exit(0);
```
koonpeng, line 417: Personally, I still prefer if rclnodejs does not install any handlers, it stills shutdown "normally" on ctrl+c for me (in linux, maybe it's different for other OS).

If we are installing handlers, I think we should at least remove `process.exit(0)`, if we shutdown all contexts, node should exit automatically when it finishes all pending promises.

koonpeng (commented):
Thanks for this, I added some of my thoughts.

> I would expect the code rclnodejs.shutdown() to shut down all existing contexts instead of only the default one. Is that intentional? Is there a will to change that?

I think it's because the API tries to follow that of rclpy.

index.js
felixdivo: That's true. I wanted to *start* documenting it. It can be extended at any time.

minggangw:
Sorry for the delay, some comments and will review the PR later, thanks!

> Some other stuff that I noticed and wanted to discuss. In [`rclnodejs/lib/context.js`](https://github.com/RobotWebTools/rclnodejs/blob/develop/lib/context.js):
> 
>     1. Should we remove `Context.shutdownDefaultContext()`? I'd argue that we want to have a `Context.tryShutdownDefaultContext()` too, but that would start to weirdly duplicate stuff. Instead, I'd suggest the following:
>        
>        * Delete `Context.shutdownDefaultContext()`, which would now become the barely longer `Context.defaultContext().shutdown()`.

+1 `Context.defaultContext().shutdown()` is more consistent 

>        * By that, we get the "try" version for free, e.g. with `Context.defaultContext().tryShutdown()`.
>        * Finally, we should add a check like `if (this === defaultContext) { defaultContext = undefined; }` to `Context.shutdown()`/`Context.tryShutdown()` anyways, as we would miss the cleanup if a user already explicitly chose to shut down the default context. This would, however, make `Context.shutdownDefaultContext()` even more redundant.

Would you please write a code snippet to show your idea? and we could discuss based on the PR.


> 
>     2. Shouldn't `defaultContext` be set to `null` instead of `undefined` to explicitly signal that it is absent and not just a typo somewhere?

Yes, 'null' is reasonable.

>     3. I would expect the code `rclnodejs.shutdown()` to shut down _all_ existing contexts instead of only the default one. Is that intentional? Is there a will to change that?

We may create multi context and only shutdown one of them at a time, so we cannot close all contexts. Please see the corresponding test case:
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/test/test-init-shutdown.js#L164-L169

>     4. Shouldn't we call `NodeOptions.defaultOptions` in [`index.js`](https://github.com/RobotWebTools/rclnodejs/blob/develop/index.js#L187) instead of reference that method? I.e. append `()`.

Would you please explain more? which one do you prefer?

index.js
minggangw: > If we are installing handlers, I think we should at least remove process.exit(0), 

We have to call `process.exit(0)`, otherwise, the process will not terminate, because the original handler doesn't get called (we changed the behaviour by installing our own handler)

Please see my comments on #720, I think we could have two ways here:

- remove the `SIGINT` handler
- propagate `SIGINT` and remove `process.exit(0)`, so we could quit as normally as if there is no handler

index.js
minggangw: I suggest we keep the naming `shutdown` to align with that of `rclpy`and simply change the current `shutdown` method's behaviour to not throw exceptions when the context has been shut down (but may still throw exceptions thrown from `rcl`).

To indicate the case that the context is already shut down, we could use `debug()` to notify, how do you think @koonpeng @felixdivo?

minggangw (commented):
Thanks for your patch, some comments below, thanks!

koonpeng:
sorry, my mistake. Removing the installed handler does indeed cause node to not exit. I think node should exit when it runs out of work in the event loop, I tested this by setting a timeout and installing a signal handler to clear it (does not call process.exit) and node exited normally. But when i try it with rclnodejs, I shutdown the context and destroy all nodes I created in the handler but node doesn't stop. I don't know if my understanding of node is wrong or if rclnodejs is leaking some things in the event loop, I will have to research into it more.

minggangw:
>  I don't know if my understanding of node is wrong or if rclnodejs is leaking some things in the event loop, 

I suspect the sub-thread hasn't stopped (the default event loop is still running), so the main thread cannot quit.
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L97-L98

index.js
felixdivo: Sounds good.

koonpeng:
Here is what I found in my investigation.

`rcl.init` creates a guard condition
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/rcl_bindings.cpp#L116

The only place I see this guard conditional used is here
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L201-L217
There is a weird check of the wait set if there is only 1 guard condition and it is truthy, if so it sets `running_` to be false, presumbly this is the `g_sigint_gc` guard conditional.

At the executor cleanup, it only runs if `running_`is true.
https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L94

So my guess is that when SIGINT is received, the guard condition gets triggered, causing the if check to pass and set `running_` to false, this happens in the C++ signal handler. Then node's SIGINT handler gets called, which does a `rclnodejs.shutdown`, but since `running_` is false, no shutdown actually occurs.

I tried commenting out the line and that seems to fix it, ctrl+c without `process.exit` in the handler properly exits node.

I don't quite understand the intention of this check, there are also several other things that seems weird to me.

https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L74-L75
here the wait set is initialized with capacity for 2 guard conditions for some reason, I guess 1 is for `g_sigint_gc` but I don't find where is the 2nd slot for.

https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L174-L187
here the guard condition is resized, I don't know which handlers `handle_manager` is tracking but it returns 0 in my testing, so the number of guard conditions get resized to 1.

https://github.com/RobotWebTools/rclnodejs/blob/d41f7e91510cd55ff672dad8f7f0be095ed8c00c/src/executor.cpp#L201-L202
Also back here again, a new guard condition is added, this happens each time in the "run loop", this guard condition is never removed, I suppose it gets discarded from the resize. Is the goal to have the `g_sigint_gc` at the end of the array? But I don't know if it makes any difference?

koonpeng:
Also do we really need a guard condition for SIGINT? I tried removing it and things still seems to work.

felixdivo:
I addressed the comments above and collected the state of this PR in the description at the top. I did change the interface of `Context` a bit, so that's something to look at maybe.

I'm somewhat overwhelmed by the discussion on the exit, as I'm not a node.js native binding guru. ;) Great discussion, but guess I'll just listen to your conclusion and implement the result if you want.

lib/context.js
```diff
@@ -16,59 +16,74 @@
...
+      this.shutdown();
     }
   }
 
-  handle() {
+  /**
+   * Get the handle referencing the internal context object. Do not modify it yourself: only pass it to *rclnodejs* functions!
+   * @return {undefined} a reference to the internal context object
+   */
+  get handle() {
```
minggangw, line 57: Please change from `context.hand()` to `context.handle` in index.js because of getter.

minggangw (commented):
Thanks for your PR, lgtm after the comments are addressed

lib/context.js
```diff
@@ -16,59 +16,74 @@
...
       defaultContext = new Context();
     }
     return defaultContext;
   }
-
-  /**
-   * Shutdown the default context.
-   * @return {undefined}
-   */
-  static shutdownDefaultContext() {
```
minggangw, line 68: Please change https://github.com/RobotWebTools/rclnodejs/blob/178c2c7068e9c1961f0db3328ab74a366803340e/index.js#L312 accordingly

minggangw:
@koonpeng thanks for your comprehensive investigation, indeed, the guard condition related codes seem unreasonable, I think one possible reason is that the guard condition was not implemented at the beginning and some value was hard code (as the code snippet you show), but the original code was not cleaned up when the new feature was added. So, it's a good opportunity to do it right, would you please update #720 to fix what you have found, thanks a lot!

index.js
felixdivo: We should change it in your PR

lib/context.js
felixdivo: Done in 84d7816

lib/context.js
felixdivo: Already done here: https://github.com/RobotWebTools/rclnodejs/pull/721/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346L315

felixdivo:
So it seems like this PR is ready to be merged. Is there some deprecation policy (I couldn't find one) or should we just add an entry to the release notes (I'd prefer that since it should be in a major release anyhow)?

minggangw:
All changes info will be added to the release note when a new version is being published.

felixdivo:
@minggangw The CI checks now pass. I improved one test to be less duplicated and such that it also checks some properties of the `Context` class.

minggangw (approved):
Thanks for the PR, congrats on your first patch! lgtm

minggangw:
@koonpeng any comments on this PR? If not, I am going to merge it :)

koonpeng (approved):
lgtm

PR #722, opened 2020-11-13
lifecycle api, types, tests & example

### LINKED ISSUE #589: LifecycleNode support in rclnodejs
rclnodejs should support lifecycle (a.k.a managed) nodes.

https://github.com/ros2/demos/blob/master/lifecycle/README.rst

wayneparrott:
This PR implements the ROS2 Lifecycle Node (managed node) as described in the [ROS2 Managed Node design](https://design.ros2.org/articles/node_lifecycle.html) and implemented by the [rclcpp lifecycle implementation](https://github.com/ros2/rclcpp/tree/master/rclcpp_lifecycle).

### Key changes:

**index.js (updated)**
* added lifecyle import
* added rclnodejs.lifecycle property - simulate namespace
* added createLifecycleNode()
* refactored common logic for node creation for use in both createNode() and createLifecycleNode()

**index.ts (updated)**
* added TypeScript declarations for createLifecyleNode()

**lifecycle.js (new)**
* includes LifecycleNode, a subclass of ShadowNode, and lifecycle support classes and interfaces

**index.d.ts (new)**
* includes TypeScript declarations for LifecycleNode and lifecycle support classes and interfaces.

**rcl_lifecycle_bindings.hpp (new)**
**rcl_lifecycle_bindings.cpp (new)**
* contains native addon C functions for interfacing with rcl_lifecycle_state_machine data-structures and api.

**lifecycle_publisher.js (new)**
* implements LifecyclePublisher - based on the rclcpp LifecyclePublisher implementation

**lifecycle_publisher.d.ts (new)**
* TypeScript declarations for LifecyclePublisher.

**publisher.js (updated)**
* refactored common publisher creation functionality for reuse by LifecyclePublisher.

**test-lifecycle.js (new)**
* unit tests for LifecycleNode and related classes

**test-lifecycle-publisher.js (new)**
* unit tests for LifecyclePublisher and related classes

**main.ts (updated)**
* added dtslint LifecycleNode cases

**lifecycle-node-example.js (new)**
* demonstrates how to use LifecycleNode, register transition callbacks and change states.

Fix #589


src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,372 @@
...
+  // each service RclHandler with a new function that is passed a pointer to the
+  // service RclHandler. The deleter function then sets the service RclHandler
+  // parameter's pointer to nullptr. All joy - no double free()!
+  auto srv_handle =
+      RclHandle::NewInstance(service, nullptr, [] { return RCL_RET_OK; });
+  RclHandle* rclHandle = RclHandle::Unwrap<RclHandle>(srv_handle);
+  RclHandle::Unwrap<RclHandle>(srv_handle)->set_deleter([rclHandle] {
+      rclHandle->set_ptr(nullptr);
+      return RCL_RET_OK;
+    });
```
wayneparrott: I would appreciate a close critique of this unit of code where I worked around an exception where free() was being called twice on the same underlying rcl unit of memory. In a nutshell, rcl provides a rcl_lifecycle_state_model that contains a state-model and also instances of Services for the lifecycle_msgs/srv/* interfaces. The LifecycleNode implementation accesses these Services from the rcl_lifecycle_state_model, creates RclHandles for each and then installs them in the node's services list. When the node is gc'ed the deleter for the RclHandle wrapping the rcl_lifecycle_state_model properly cleans up it's memory and releases the services it manages. The RclHandles wrapping each of the state_model services are calling free() on their service pointer which has already been freed by it's containing rcl_lifecycle_state_model. This is causing a segfault.

To avoid this 2nd free() I'm hacking each service RclHandler deleter such that it sets it's pointer to a service to null. Doing this prevents RclHandle#Reset() from calling free(). No more **double free()**.

Hope this makes sense.

index.js
```diff
@@ -424,3 +429,30 @@ const TimeSource = require('./lib/time_source.js');
 rcl.TimeSource = TimeSource;
 
 inherits(rclnodejs.ShadowNode, Node);
+
+function _createNode(
+  nodeName,
+  namespace = '',
+  context = Context.defaultContext(),
+  options = NodeOptions.defaultOptions,
+  nodeClazz
```
minggangw: nodeClass? I googled `clazz` and found it was used in Java commonly :)

lib/lifecycle.js
```diff
@@ -0,0 +1,748 @@
...
+  isError() {
+    return this.value === CallbackReturnCode.ERROR;
+  }
+
+  /**
+   * A predicate to test if an error message has been assigned.
+   * @returns {boolean} Return true if an error message has been assigned; otherwise return false.
+   */
+  hasErrorMsg() {
+    return this.isSuccess() ? false : !!this._errorMsg;
```
minggangw: How about `return !this.isSuccess() && this._errorMsg;`

src/rcl_lifecycle_bindings.hpp
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLNODEJS_RCL_LIFECYCLE_BINDINGS_HPP_
+#define RCLNODEJS_RCL_LIFECYCLE_BINDINGS_HPP_
+
+#include <nan.h>
+#include <rcl/rcl.h>
```
minggangw: these headers can be removed

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,372 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rcl_lifecycle_bindings.hpp"
+
+#include <rcl/error_handling.h>
+#include <rcl/rcl.h>
+
+#include <iostream>
```
minggangw: I think we don't need this.

src/rcl_lifecycle_bindings.cpp
minggangw: This is a new scenario of `RclHandle`, currently, the pointer was freed by default when calling `Reset()`. I suggest we could extend the `RclHandle` class to make the code consistent

For the declaration:
```c
  // |handle| will be freed if |will_delete| is true.
  static v8::Local<v8::Object> NewInstance(
      void* handle, RclHandle* parent = nullptr,
      std::function<int()> deleter = [] { return 0; },
      will_delete = true)
```

For the implementation:
```c
if (will_delete_)
  free(pointer_);
```

than we could call like
```c
 auto srv_handle =
      RclHandle::NewInstance(service, nullptr, std::function<int()>(), false);
```

Also, I notice that we have https://github.com/RobotWebTools/rclnodejs/blob/711072d689acc7221cbb8e289ca552db2c423c4b/src/rcl_handle.cpp#L81-L86, not sure if we could leverage that.

minggangw (commented):
Thanks for implementing this HUGE pull request and I look through it quickly, generally looks good! As I don't understand lifecycle comprehensively, I'd like to invite @koonpeng to review it, thanks!

koonpeng:
this is quite a large PR, I may need some time to review this :sweat_smile:. I don't have experience with lifecycle nodes as well so I will just try to review based on my knowledge of rcl.

wayneparrott:
@koonpeng @minggangw et al, please ping me if I can be of assistance during the review.

At work I usually start reviews by asking the dev how comfortable he/she is with a specific unit of code, are there any hot spots or overly complex areas that he has concerns. With that I'll share 2 areas that need discussion:

1. I've already highlighted that the implementation makes 4 Services created by the rcl lifecycle api visible to rclnodejs. A work-around of the RclHandler Reset behavior was implemented to avoid double free() error of the Service ptr.  

edited @wayneparrott - Please hold off on further review. While describing design issue #2 below I identified a simplification which I will submit an update later today.

~~2. The lifecycle.js module requires an `initialize()` function to be called before it is runnable. I'm not fully comfortable with this. This requirement arises because of a dependency cycle that arises because I was trying to use lifecycle state machine constants defined in the lifecycle [State and Transition messages](https://github.com/ros2/rcl_interfaces/tree/master/lifecycle_msgs) rather than hard-code those same constants in JS code.  An alternate approach is to hard-code the lifecycle state machine constants defined in the lifecyle State and Transition messages. **I'm not opposed to the latter approach.** But wanted to get the insight of others before doing so.~~

~~tldr;
Here's the details behind this and an alternate approach:~~

~~The `State` and `Transition` message definitions include the state id constants used by the lifecycle state machine. The implementation loads instances of these message types at runtime in order to reference their state machine constants. This dependency leads to the Lifecycle module not being runnable until message files have been generated which happens at rclnodejs installation and at `rlcnodejs.init()` time. Here's where it gets tricky, index.js loads the lifecycle module and exposes it through a new public lifecycle property, i.e., `rclnodejs.lifecycle`. To work around this chicken-egg dependency the lifecycle module has an initialize() function that must be called after calling rclnodejs.init(). At this time, lifecycle.initialize() is called implicitly as a part of rclnodejs.createLifecycleNode().~~

~~Summary of the dependency:  rclnodejs -> lifecycle module -> generated messages -> rclnodejs.init()~~

~~**The lifecycle module dependency upon the State and Transition messages could be eliminated by hard-coding their state machine constants in JS code rather than pulling them from these 2 messages. I was hoping to avoid hardcoding the constants. But as a consequence, the lifecycle module has to wait for rclnodejs.init() to complete before it can safely load the 2 messages.**~~

~~So before going too deep into the review, let's decide if the current design with dependencies on the message constants is acceptable or if we should work to remove the dependency on the State and Transition messages, i.e., hard code the  constants. **At this time I'm very open to hard coding the constants. Thoughts?**
tldr;~~

lib/lifecycle.js
wayneparrott: +1

wayneparrott:
The latest commit should be the one. It includes @minggangw earlier suggestions; especially see rclhandler.cpp free_ptr_. In lifecycle.js I switch to using lazy initialization of a couple of message interfaces which I feel is a good simplification from my initial commit.

src/rcl_lifecycle_bindings.cpp
koonpeng: I suggest to instead make the `deleter`'s job to free the memory, the default delete will call free but you can pass in a custom deleter to perform other actions.

src/rcl_handle.cpp
```diff
@@ -86,7 +86,8 @@ NAN_METHOD(RclHandle::Dismiss) {
 }
 
 v8::Local<v8::Object> RclHandle::NewInstance(void* handle, RclHandle* parent,
-                                             std::function<int()> deleter) {
+                                             std::function<int()> deleter,
```
koonpeng: I suggest to instead make the `deleter`'s job to free the memory, the default delete will call free but you can pass in a custom deleter to perform other actions. This follows closer to how `std::shared_ptr` and other v8 functions that takes in a custom deleter works.

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "lifecycle_msgs/srv/get_state.h"
+
+
+namespace rclnodejs {
+
+static v8::Local<v8::Object> wrapState(const rcl_lifecycle_state_t* state) {
+  v8::Local<v8::Object> jsState = Nan::New<v8::Object>();
+  Nan::Set(jsState, Nan::New("id").ToLocalChecked(), Nan::New(state->id));
+  Nan::Set(jsState, Nan::New("label").ToLocalChecked(),
+           Nan::New(state->label).ToLocalChecked());
```
koonpeng, line 40: I don't know much about native c++ modules, would it be better if we use v8 accesors to bind these properties? https://v8.dev/docs/embed#accessing-dynamic-variables and https://github.com/nodejs/nan/blob/master/doc/methods.md#api_nan_set_accessor.

We might also want to make them readonly.

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,363 @@
...
+  return jsState;
+}
+
+static v8::Local<v8::Object> wrapTransition(
+    const rcl_lifecycle_transition_t* transition) {
+  v8::Local<v8::Object> jsTransition = Nan::New<v8::Object>();
+  Nan::Set(jsTransition, Nan::New("id").ToLocalChecked(),
+           Nan::New(transition->id));
+  Nan::Set(jsTransition, Nan::New("label").ToLocalChecked(),
+           Nan::New(transition->label).ToLocalChecked());
```
koonpeng, line 50: See above comment in `SetAccessor`.

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,363 @@
...
+  rcl_lifecycle_state_machine_t* state_machine =
+      reinterpret_cast<rcl_lifecycle_state_machine_t*>(
+          state_machine_handle->ptr());
+
+  std::string lifecycle_srv_field_name(*Nan::Utf8String(info[2]));
+
+  std::string service_name;
+  rcl_service_t* service;
+  if (lifecycle_srv_field_name.compare("srv_get_state") == 0) {
+    service_name = "~/get_state";
```
koonpeng: Should we be using `rcl_service_get_service_name` instead of hardcoding this?

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,363 @@
...
+    service_name = "~/get_available_states";
+    service = &state_machine->com_interface.srv_get_available_states;
+  } else if (lifecycle_srv_field_name.compare(
+                 "srv_get_available_transitions") == 0) {
+    service_name = "~/get_available_transitions";
+    service = &state_machine->com_interface.srv_get_available_transitions;
+  } else if (lifecycle_srv_field_name.compare("srv_change_state") == 0) {
+    service_name = "~/change_state";
+    service = &state_machine->com_interface.srv_change_state;
+  }
```
koonpeng, line 224: missing find `else` so we may end up creating a `RclHandle` to a random access or nullptr (does c++ initialize pointers?).

src/rcl_lifecycle_bindings.cpp
```diff
@@ -0,0 +1,363 @@
...
+  rcl_lifecycle_state_machine_t* state_machine =
+      reinterpret_cast<rcl_lifecycle_state_machine_t*>(
+          state_machine_handle->ptr());
+  std::string transition_label(*Nan::Utf8String(info[1]));
+
+  bool publish = true;
+  rcl_ret_t ret = rcl_lifecycle_trigger_transition_by_label(
+      state_machine, transition_label.c_str(), publish);
+
+  if (ret == RCL_RET_OK) {
```
koonpeng: throw on error?

lib/lifecycle.js
```diff
@@ -0,0 +1,728 @@
...
+ * @typedef SerializedState
+ * @type {object}
+ * @property {number} id - code identifying the type of this state.
+ * @property {string} label - readable name of this state.
+ */
+
+/**
+ * The state of the lifecycle state model.
+ */
+class State {
```
koonpeng, line 56: We are already creating a new object with `wrapState` in the c++ side, we could avoid creating redundant objects by either using the object from c++, or bind the values on c++ side and create a copy of it here to avoid unintended modifications by the user.

lib/lifecycle.js
```diff
@@ -0,0 +1,728 @@
...
+  constructor(id, label) {
+    this.id = id;
+    this.label = label;
+  }
+
+  /**
+   * Create an object representation of state properties.
+   * @return {SerializedState} The object.
+   */
+  asMessage() {
```
koonpeng: Doesn't this return the same thing except it removes the prototype chain?

lib/lifecycle.js
```diff
@@ -0,0 +1,728 @@
...
+    this.startState = startState;
+    this.goalState = goalState;
+  }
+
+  /**
+   * Create an object representation of the transitionDescriptons properties.
+   * 
+   * @returns {object} The object representation.
+   */
+  asMessage() {
```
koonpeng: Doesn't this return the same thing except it removes the prototype chain?

koonpeng (commented):
Echoing back to `SetAccessors`, I wonder if it's possible to bind all the rcl structs to js objects.

src/rcl_lifecycle_bindings.cpp
wayneparrott: +1, I completely missed this api and hardcoded as my fallback plan. Will make this change.

src/rcl_lifecycle_bindings.cpp
wayneparrott: +1. Added the following snippet:
```
THROW_ERROR_IF_EQUAL(
                   nullptr,
                   service,
                   "Service not found.");
```

src/rcl_lifecycle_bindings.cpp
wayneparrott: +1. Using this macro now throughout the file.
```
  THROW_ERROR_IF_NOT_EQUAL(RCL_RET_OK,
                           rcl_lifecycle_trigger_transition_by_id(
                              state_machine, transition_id, publish),
                           rcl_get_error_string().str);
```

lib/lifecycle.js
wayneparrott: +1. Removed asMessage() methods from State, Transition and TransitionDescription.

lib/lifecycle.js
wayneparrott: +1. see above.

src/rcl_handle.cpp
wayneparrott: @koonpeng @minggangw I can see the merit to your recommendation to delegate managing ptr memory to a deleter function. Yet I feel making such a change is beyond the scope of this PR due to the large impact it would have across the addon codebase. My preference is to continue with my original technique/hack, i.e., where my deleter function sets the rclHandle's ptr to null, and open a separate issue for how RclHandle handles free it's ptr.

index.js
wayneparrott: Renaming to `nodeClass`. From my zillion years developing with Java I used 'zz' without giving it real thought.

src/rcl_lifecycle_bindings.cpp
wayneparrott: My initial thought regarding your suggestion to bind JS objects to C++ object wrappers on rcl C structs is a more object-oriented approach and a departure from the current functional style addon api design. I'm hesitant to embark on a new approach without broader discussion and agreement. It seems like a slippery slope that could quickly get very complicated when generally applied.

I realize my impl dynamically creates JS data-transfer objects and copies data into them rather than directly referencing the C data from JS objects. In this case is seemed like an OK tradeoff of a little extra memory per state/transition data for a simple implementation. Maybe this is the wrong attitude given this code could be running on a smaller system, e.g., raspi. 

My preference is to continue with the current api design. But I'm open (team player) to a general change towards an object-oriented wrapping approach. Before we go there, consider the potential implementation complexity over our functional model. My expectation is adopting an OO approach will result in creating parallel C++ hierarchies wrapping rcl C structs and api with corresponding JS bindings. So in the lifecycle case that could involve creating C++ representations for Lifecycle StateMachine, State, Transition, TransitionDescription and Service.

@minggangw @koonpeng  Thoughts?

koonpeng: I don't mind the current design, the only small problem I see is that some functions are creating a temporary object which are immediately discarded on the js side. But I don't think these functions are called often enough that it matters

Currently these are only simple plain data objects, but if we want to add methods to them in the future, it could create quite a lot of extra complexity that might not be worth the trouble.

wayneparrott: Agree. Thx for the feedback.

src/rcl_handle.cpp
minggangw: +1 We'd better focus on the lifecycle feature within this PR, considering we have already had a lot of changes. @wayneparrott please go-ahead to open an issue, as @koonpeng said we should have a completely user-defined deleter.

wayneparrott: Opened issue #728

src/rcl_lifecycle_bindings.cpp
wayneparrott: see comments above.

lib/lifecycle.js
wayneparrott: See previous comments.

src/rcl_lifecycle_bindings.cpp
wayneparrott: see issue #728

minggangw (approved):
I have no further comments, lgtm, and thanks for implementing this important feature, great progress!

minggangw:
I think we could move forward, @koonpeng any comments? @wayneparrott please merge it when it's good to go (soft reminder: please select "Squash and merge" option)

koonpeng (approved):
LGTM!

PR #737, opened 2020-11-21
Simplify rclnodejs/index.js

### LINKED ISSUE #736: rclnodejs/index.js could be simplified
The file `rclnodejs/index.js` could be simplified using `async function`s, making it easier to maintain. Other ways to simplify it would be array handling (with `Array.some()` and `Array.every()`) and replacing `let` with `const` here and there.

felixdivo:
Fixes #736.

### DIFF
--- a/index.js
+++ b/index.js
@@ -347,22 +347,15 @@ let rcl = {
    * this folder will be overwritten after the execution.
    * @return {Promise<undefined>} A Promise.
    */
-  regenerateAll() {
+  async regenerateAll() {
     // This will trigger to regenerate all the JS structs used for messages and services,
     // to overwrite the existing ones although they have been created.
-    debug('Begin regeneration of JavaScript code from ROS IDL files.');
-    return new Promise((resolve, reject) => {
-      generator
-        .generateAll(true)
-        .then(() => {
-          tsdGenerator.generateAll(); // create interfaces.d.ts
-          debug('Finish regeneration.');
-          resolve();
-        })
-        .catch((e) => {
-          reject(e);
-        });
-    });
+    debug('Begin regeneration of JavaScript code from ROS IDL files...');
+
+    await generator.generateAll(true);
+    await tsdGenerator.generateAll(); // create interfaces.d.ts
+
+    debug('Finished regeneration.');
   },
 
   /**
--- a/index.js
+++ b/index.js
@@ -355,7 +355,7 @@ let rcl = {
     // to overwrite the existing ones although they have been created.
     debug('Begin regeneration of JavaScript code from ROS IDL files...');
 
-    # generate the messages and type declarations, which must be done in sequence
+    // generate the messages and type declarations, which must be done in sequence
     await generator.generateAll(true);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -344,22 +350,15 @@ let rcl = {
...
-          debug('Finish regeneration.');
-          resolve();
-        })
-        .catch((e) => {
-          reject(e);
-        });
-    });
+    debug('Begin regeneration of JavaScript code from ROS IDL files...');
+
+    await generator.generateAll(true);
```
felixdivo, line 359: Does this need to happen in sequence or could this also be done "concurrently" with `Promise.all([...])`?

minggangw: The ts generator depends on the JS files generated by `generator.generateAll(true)`, @wayneparrott please help to take a look, thanks!

wayneparrott: @minggangw @felixdivo Yes, message generation is a 2 phase process: 1) generate js message files, 2) generate the interfaces.d.ts declaration file from the js message files.

felixdivo: Okay, thanks. Then this PR should be done. I'll just add a comment about the above.

felixdivo:
```suggestion
    # generate the messages and type declarations, which must be done in sequence
    await generator.generateAll(true);
```

felixdivo:
I don't get why AppVeyor is failing though. But maybe it will go away now. Let's see.

index.js
```diff
@@ -344,22 +350,16 @@ let rcl = {
...
-          debug('Finish regeneration.');
-          resolve();
-        })
-        .catch((e) => {
-          reject(e);
-        });
-    });
+    debug('Begin regeneration of JavaScript code from ROS IDL files...');
+
+    # generate the messages and type declarations, which must be done in sequence
```
minggangw: I think this line causes the failure on ci :)

felixdivo: Oh WTF, too much Python ... :sweat_smile:

felixdivo:
> I don't get why AppVeyor is failing though. But maybe it will go away now. Let's see.

It didn't. [This](https://ci.appveyor.com/project/minggangw/rclnodejs/builds/36461095#L475) war the failing test I meant (@minggangw ;)). Seems like `node ./scripts/compile_tests.js` fails. I can't see why this would be cause by my changes. It appears that a few lines above, this shows up:

```
> node scripts/generate_messages.js
Start JavaScript message generation...
Incomplete service: test_interface_files.srv.Arrays.
Incomplete service: test_interface_files.srv.BasicTypes.
Incomplete service: test_interface_files.srv.Empty.
Generation complete.
```

But that also happens on the main branch on [AppVeyor](https://ci.appveyor.com/project/minggangw/rclnodejs/builds/36440411#L436) and probably also Travis CI.

minggangw:
I've re-triggered the appveyor,  my experience is that the appveyor is not stable sometimes and you have to retry it. Hope it will pass this time and I will merge this PR once it's done, thanks!

minggangw (approved):
Thanks for making the code more readable, lgtm

PR #743, opened 2020-11-28
Use specified deleter for RclHandle

### LINKED ISSUE #728: Improve RclHandle ptr free() strategy
@koonpeng recently proposed redesigning the RclHandle to make the deleter lambda function responsible for `free()`ing the RclHandle memory pointer rather than `free()`ing its ptr directly as it does now. 

tldr;
The [RclHandle](https://github.com/RobotWebTools/rclnodejs/blob/develop/src/rcl_handle.cpp) `Reset()` method divides clean up duties between its self and a custom deleter lambda function provided during instance initialization. The deleter function performs custom delete activities on the RclHandle's ptr. While the RclHandle directly `free()`s its memory ptr. Recently this split in responsibilities resulted in a double `free()` error where the RclHandle ptr was free()'ed by a parent RclHandle and then again freed by its RclHandle wrapper.. The resolution required a creative hack using a custom deleter function that tricked the RclHandler to bypass calling free() on its ptr. This scenario revealed the need to externalize all of the clean up of the RclHandler memory ptr. The RclHandle deleter function is a natural candidate for the memory ptr free() responsibility.

minggangw:
This patch will enable RclHandle to use a specified deleter to free
the resources, which means the pointer passed when contructing the
RclHandle object will not be freed by default.

After this patch, caller should be responsible to manage the memory.

Fix #728

### DIFF
--- a/src/rcl_handle.cpp
+++ b/src/rcl_handle.cpp
@@ -85,8 +85,8 @@ NAN_METHOD(RclHandle::Dismiss) {
   info.GetReturnValue().Set(Nan::Undefined());
 }
 
-v8::Local<v8::Object> RclHandle::NewInstance(void* handle, RclHandle* parent,
-                                             std::function<int()> deleter) {
+v8::Local<v8::Object> RclHandle::NewInstance(
+    void* handle, RclHandle* parent, std::function<void(void*)> deleter) {
   Nan::EscapableHandleScope scope;
 
   v8::Local<v8::Function> cons = Nan::New<v8::Function>(constructor);
@@ -123,12 +123,7 @@ void RclHandle::Reset() {
     child->Reset();
   }
 
-  if (deleter_ && deleter_() != RCL_RET_OK) {
-    Nan::ThrowError(rcl_get_error_string().str);
-    rcl_reset_error();
-  }
-
-  free(pointer_);
+  if (deleter_) deleter_(pointer_);
 
   pointer_ = nullptr;
   children_.clear();
--- a/src/rcl_lifecycle_bindings.cpp
+++ b/src/rcl_lifecycle_bindings.cpp
@@ -21,15 +21,14 @@
 #include <string>
 #include <vector>
 
-#include "macros.hpp"
-#include "rcl_handle.hpp"
-#include "rcl_utilities.hpp"
 #include "lifecycle_msgs/msg/transition_event.h"
 #include "lifecycle_msgs/srv/change_state.h"
 #include "lifecycle_msgs/srv/get_available_states.h"
 #include "lifecycle_msgs/srv/get_available_transitions.h"
 #include "lifecycle_msgs/srv/get_state.h"
-
+#include "macros.hpp"
+#include "rcl_handle.hpp"
+#include "rcl_utilities.hpp"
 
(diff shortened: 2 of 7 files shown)


minggangw:
@wayneparrott @koonpeng PTAL, thanks!

src/rcl_lifecycle_bindings.cpp
```diff
@@ -223,24 +224,8 @@ NAN_METHOD(GetLifecycleSrvNameAndHandle) {
...
-  // Note: lifecycle Services are created and managed by their
-  // rcl_lifecycle_state_machine. Thus we must not manually
-  // free the lifecycle_state_machine's service pointers.
-  // To accomplish this we create srv_handle instances with
-  // the free_ptr parameter of false. Failing to do this results
-  // in a double free() error.
-  auto srv_handle =
-      RclHandle::NewInstance(service, nullptr, [] { return RCL_RET_OK; });
+  auto srv_handle = RclHandle::NewInstance(service, nullptr, nullptr);
   RclHandle* rclHandle = RclHandle::Unwrap<RclHandle>(srv_handle);
```
wayneparrott: With the new deleter design I think this line is no longer needed.
```
 RclHandle* rclHandle = RclHandle::Unwrap<RclHandle>(srv_handle);
```

wayneparrott (approved):
LGTM. My C coding skills are WIP. Thus, recommend giving more weight to any @koonpeng feedback. Thx.

src/rcl_lifecycle_bindings.cpp
minggangw: Yes, and I want to get some meaningful comments back here:

  // Note: lifecycle Services are created and managed by their
  // rcl_lifecycle_state_machine. Thus we must not manually
  // free the lifecycle_state_machine's service pointers.

src/rcl_handle.cpp
```diff
@@ -85,8 +85,8 @@ NAN_METHOD(RclHandle::Dismiss) {
   info.GetReturnValue().Set(Nan::Undefined());
 }
 
-v8::Local<v8::Object> RclHandle::NewInstance(void* handle, RclHandle* parent,
-                                             std::function<int()> deleter) {
+v8::Local<v8::Object> RclHandle::NewInstance(
+    void* handle, RclHandle* parent, std::function<void(void*)> deleter) {
```
koonpeng, line 89: Should we have the default deleter to `free` the pointer to simplify usage for the most common case?

On a side note, I recently found out that `free` on a c++ `new` allocated pointer is undefined behavior, similar for c++ `delete` on a `malloc` allocated pointer. This is another reason why moving the work of deallocation to the deleter function is a good idea, the old implementation will call `free` regardless of the pointer being `malloc` or `new` allocated.

koonpeng (approved):
LGTM, I just have a minor thought of a default deleter function.

src/rcl_handle.cpp
minggangw: As you said, we cannot assume whether the `ptr` is allocated by `malloc` or `new`, so letting the caller decide seems reasonable. Although, we have to repeat `free` in most cases, any thoughts?

PR #744, opened 2020-12-01
Add prettierrc

koonpeng:
This adds a `.prettierrc.yml` file so that prettierrc config file outside the project will not get mistakenly loaded when it goes through husky/lint-staged. I also did a run of prettier and clang-format on the current files, feel free to revert the formatting commits if they are undesired.

Most of the changes are just formatting, only `.prettierrc.yml` and `package.json` changes are relevalent.


koonpeng:
The prettier formatting is conflicting with the eslint settings, even though the eslint config extends eslint-config-prettier, it still defines the incompatible rules so that overwrites them.

Do we use eslint or prettier to perform formatting? And should I change the formatting according to the current eslint or prettier rules?

minggangw:
> Do we use eslint or prettier to perform formatting? 

Personally, I prefer prettier.

> And should I change the formatting according to the current eslint or prettier rules?

+1 I'd like to format the whole codebase.

I found a article to solve the conflict that I think we could follow https://blog.theodo.com/2019/08/empower-your-dev-environment-with-eslint-prettier-and-editorconfig-with-no-conflicts/

1. Write .eslintrc.yml with:
```json
{
  "extends": ["eslint:recommended", "prettier"],
  "env": {
    "es6": true,
    "node": true
  },
  "rules": {
    "prettier/prettier": "error"
  },
  "plugins": [
    "prettier"
  ]
}
```
2. and add `eslint-plugin-prettier` as devDepenencies. I test locally and it works well :)

minggangw:
I changed our `.eslintrc.yml` to 
```json
env:
  node: true
  es6: true
globals:
  every: true
  after: true
  constantly: true
ignorePatterns: ["generated/"]
rules:
  prettier/prettier: "error"
extends: ["eslint:recommended"]
plugins: ["prettier"]
parserOptions:
  ecmaVersion: 2018
```

koonpeng:
there were some differences between the eslint rules and prettier, for example the eslint rules had `max-len: ["error", {"code": 120, "tabWidth": 2, "ignoreComments": true}]`, iirc prettier's default line length is 80 and we did not change it in args. Should I respect the prettier settings or the eslint rules?

minggangw:
I suggest we could follow the prettier rules only and remove the current ones, it seems there is always kind of possibility that mixing rules of prettier and eslint leads to conflicts.

minggangw (approved):
lgtm, thanks for enabling this!

PR #746, opened 2020-12-03
find ros packages from ament index

koonpeng:
When in a path that has ament executed, use the ament index to obtain a list of ros definition files instead of recursively scanning the directory.

This improves the message generation speed, on my pc, i get

old find packages
```
real    0m6.069s
user    0m13.387s
sys     0m3.757s
```

using ament index
```
real    0m2.189s
user    0m13.449s
sys     0m3.722s
```


minggangw:
Wow, that accelerates a lot!

minggangw (approved):
Thanks for making the performance better! LGTM!

PR #749, opened 2020-12-08
fix tests in foxy/ubuntu-20.04

koonpeng:
I'm not sure if I'm the only one having this problem, the tests are failing for me in foxy/ubuntu-20.04. I actually saw in problem before in other projects and it is due to fastRTPS not correctly exported some cmake files. The simple workaround is to add AMENT_PREFIX_PATH to CMAKE_PREFIX_PATH.

I also try to simplify the compiling step by using colcon and sourcing the output instead of copying files around.

Edit: Actually all the previous build have been failing but it seems tests still passes
```
> rclnodejs@0.17.0 test /root/rclnodejs
> node ./scripts/compile_tests.js && node --expose-gc ./scripts/run_test.js && npm run dtslint

Starting >>> cpp_nodes

Starting >>> rclnodejs_test_msgs

--- stderr: cpp_nodes
/root/rclnodejs/test/cpp/add_two_ints_client.cpp: In function â€˜example_interfaces::srv::AddTwoInts_Response_<std::allocator<void> >::SharedPtr send_request(rclcpp::Node::SharedPtr, rclcpp::Client<example_interfaces::srv::AddTwoInts>::SharedPtr, example_interfaces::srv::AddTwoInts_Request_<std::allocator<void> >::SharedPtr)â€™:
/root/rclnodejs/test/cpp/add_two_ints_client.cpp:48:23: error: â€˜rclcpp::executor::FutureReturnCodeâ€™ has not been declared
   48 |     rclcpp::executor::FutureReturnCode::SUCCESS)
      |                       ^~~~~~~~~~~~~~~~
/root/rclnodejs/test/cpp/add_two_ints_client.cpp: In function â€˜int main(int, char**)â€™:
/root/rclnodejs/test/cpp/add_two_ints_client.cpp:94:23: error: â€˜rclcpp::executor::FutureReturnCodeâ€™ has not been declared
   94 |     rclcpp::executor::FutureReturnCode::SUCCESS)
      |                       ^~~~~~~~~~~~~~~~
/root/rclnodejs/test/cpp/add_two_ints_client.cpp: In function â€˜example_interfaces::srv::AddTwoInts_Response_<std::allocator<void> >::SharedPtr send_request(rclcpp::Node::SharedPtr, rclcpp::Client<example_interfaces::srv::AddTwoInts>::SharedPtr, example_interfaces::srv::AddTwoInts_Request_<std::allocator<void> >::SharedPtr)â€™:
/root/rclnodejs/test/cpp/add_two_ints_client.cpp:54:1: warning: control reaches end of non-void function [-Wreturn-type]
   54 | }
      | ^
make[2]: *** [CMakeFiles/add_two_ints_client.dir/build.make:63: CMakeFiles/add_two_ints_client.dir/add_two_ints_client.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:84: CMakeFiles/add_two_ints_client.dir/all] Error 2
make: *** [Makefile:95: all] Error 2
---
Failed   <<< cpp_nodes [7.06s, exited with code 2]
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
         - ./node_modules
     # Test
     - run: npm install istanbul coveralls
-    - run: source ~/ros2_install/ros2-osx/local_setup.bash && export OPENSSL_ROOT_DIR="/usr/local/opt/openssl" && node scripts/compile_tests.js && node --expose-gc ./node_modules/.bin/istanbul cover ./scripts/run_test.js --report lcovonly && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
+    - run: source ~/ros2_install/ros2-osx/local_setup.bash && export OPENSSL_ROOT_DIR="/usr/local/opt/openssl" && node scripts/compile_tests.js && . install/setup.bash && node --expose-gc ./node_modules/.bin/istanbul cover ./scripts/run_test.js --report lcovonly && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
     # Teardown
     - run: find $HOME/Library/Developer/Xcode/DerivedData -name '*.xcactivitylog' -exec cp {} $CIRCLE_ARTIFACTS/xcactivitylog \; || true
     # Save test results
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,8 +28,8 @@ before_build:
   - appveyor DownloadFile https://github.com/ros2/choco-packages/releases/download/2019-10-24/tinyxml2.6.0.0.nupkg
   - appveyor DownloadFile https://github.com/ros2/choco-packages/releases/download/2019-10-24/log4cxx.0.10.0.nupkg
   - choco install -y -s c:\download\ asio bullet cunit eigen tinyxml-usestl tinyxml2 log4cxx
-  - appveyor DownloadFile https://github.com/ros2/ros2/releases/download/release-foxy-20201009/ros2-foxy-20201009-windows-release-amd64.zip
-  - 7z x -y "c:\download\ros2-foxy-20201009-windows-release-amd64.zip" -o"c:\" > nul
+  - appveyor DownloadFile https://ci.ros2.org/view/packaging/job/packaging_windows/lastSuccessfulBuild/artifact/ws/ros2-package-windows-AMD64.zip
+  - 7z x -y "c:\download\ros2-package-windows-AMD64.zip" -o"c:\" > nul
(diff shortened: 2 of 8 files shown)


koonpeng:
The windows ci is using a ci build of ros2 which mistakenly exports local directories.
```cmake
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;C:/ci/ws/install/include"
```
This causes `cpp_nodes` to fail, I tried with a release build of ros2 and it works. The test still passes because it seems `cpp_nodes` is not used in windows tests. Should I use the ci build and exclude `cpp_nodes` from the build, or switch to a release build? The downside of the release build is that we need to update it when new ros version gets released.

minggangw:
I prefer we use the [rolling-distribution](https://index.ros.org/doc/ros2/Releases/#rolling-distribution) for `develop` branch, meanwhile, stable release matches its corresponding brach of `rclnodejs`, e.g. [foxy-fitzroy](https://github.com/RobotWebTools/rclnodejs/tree/foxy-fitzroy) uses [ROS 2 Foxy Fitzroy - Patch Release 3](https://github.com/ros2/ros2/releases/tag/release-foxy-20201009) for ci. Thus, we could track the latest changes and fix the problem before it's officially released on `develop` branch.

But the rolling-distribution (or nightly build) is not stable although it's called `stable`. I ever met similar issues before, and the problems often got fixed in the next package release. I think we could exclude `cpp_nodes` because those tests that use `cpp_nodes` are not running any more, or we could wait a few days until the next ROS2 package (https://ci.ros2.org/view/packaging/job/packaging_windows/lastSuccessfulBuild/artifact/ws/ros2-package-windows-AMD64.zip) is ready.

koonpeng:
I went back a few weeks and they all have the same error
https://ci.appveyor.com/project/minggangw/rclnodejs/builds/36303858
```
CMake Error in CMakeLists.txt:
  Imported target "rclcpp::rclcpp" includes non-existent path
    "C:/ci/ws/install/include"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
```
maybe it's a unreported bug?

koonpeng:
made an issue https://github.com/ros2/ros2/issues/1068

minggangw:
I suspect it was misconfigured, let's wait for the reply from the upstream.

minggangw:
I notice there is a new Window ROS2 package generated (https://ci.ros2.org/view/packaging/job/packaging_windows/1960/), so the problem may have been fixed.

koonpeng:
It looks like the issue still isn't fixed :disappointed:

Issue #753, opened 2020-12-14
Propose making Node extendable (i.e, create Node subclasses)


wayneparrott:
This is a proposal to slightly refactor rclnodejs such that it is possible to create Node subclasses without having to add a factory function to rcl (index.js). Included in this proposal is a requirement to maintain the current rcl api as is.

tldr;
In rclnodejs the rcl module (see index.js) provides2  factory methods, one for creating basic Node instances and one for creating LifecycleNode instances. It is not possible to create instances of user-defined Node subclasses without adding additional factory methods to rcl. Over the past year of my use of rclnodejs there have been several times where I felt that creating a Node subclass would have simplified the organization of my app code, i.e., package custom code with in a node subclass. The need was not general enough to warrant adding a 1-off factory function to rcl. I encountered this limitation again recently when I implemented LifecycleNode. And I have been a little jealous at seeing that others were able to subclass Node in the rclcpp for similar purposes. **With that, I'm proposing an enhancement to the rclnodejs design such that Node subclasses can be created directly. The enhancement will not change any current public api.**

I have implemented a PR that I would like to submit if this idea is of broader interest.

I've implemented a PR that does this by:
0. at a high-level, the developer directly creates a Node using it's constructor. He can then directly spin(), stop() & destroy() a node via it's api. 
1. expose Node constructor(), includes moving handler logic from rcl.createNode() to the constructor 
2. make Context responsible for managing it's Nodes, i.e., Context contains refs to it's nodes. Moved rcl context->nodes[]mapping to Contetx as a static property. Calling context.shutdown() now destroys the context's nodes. 
3. created ShadowNode <- Node <- LifecycleNode hierarchy
4. expose spin(), spinOnce(), stop() as public Node methods
5. all changes are backward compatible with rcl api, e.g., createNode(), spin(), shutdown(), isShutdown(). But I did mark these functions as deprecated (preliminary).

Here's an example of using this new design:
```
import * as rclnodejs from ('rclnodejs');

class ExampleNode extends rclnodejs.Node {
  constructor() {
    // use remapping to rename at runtime if needed
    super('ExampleNode');
    this._publisher = this.createPublisher('Example_Publisher', 'example');
  }

  publish(msg) {
    // preprocess msg here
    this._publisher.publish(msg);
  }

 // setup custom node state here, an example might be broadcasting robot state, and then spin() 

  destroy() {
   //do some custom cleanup
   super.destroy();
}

async function main() {
[...]

PR #754, opened 2020-12-14
Support creation of Node subclasses

wayneparrott:
Added ability to create Node subclasses. 
This PR maintains the public 0.17 api.

#753

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -150,7 +150,7 @@ class Context {
     if (this.isShutdown()) return;
 
     // shutdown and remove all nodes
-    for (const node of this._nodes) {
+    for (const node of this.nodes) {
       node.destroy();
     }
 
@@ -177,7 +177,7 @@ class Context {
    * @throws {Error} If there is a problem shutting down the context.
    */
   tryShutdown() {
-    if (this.isValid()) {
+    if (this.isInitialized()) {
       this.shutdown();
     }
   }
--- a/lib/context.js
+++ b/lib/context.js
@@ -23,8 +23,29 @@ let defaultContext = null;
  * @hideconstructor
  */
 class Context {
-  constructor() {
+  constructor(argv = process.argv) {
     this._handle = rclnodejs.createContext();
+    this._nodes = [];
+    this._isShutdown = false;
+
+    rclnodejs.init(this._handle, argv);
+  }
+
+  get nodes() {
+    return Array.from(this._nodes);
+  }
+
+  addNode(node) {
+    if (this.isShutdown()) {
+      throw new Error('Can not add a Node to a Context that is shutdown');
+    }
+
+    if (this._nodes.includes(node)) {
+      // do nothing
+      return;
+    }
+
+    this._nodes.push(node);
   }
 
   /**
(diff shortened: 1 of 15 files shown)


lib/context.js
```diff
@@ -19,64 +19,201 @@ const rclnodejs = require('bindings')('rclnodejs');
...
+ *                                     |
+ *         ----------------------------
+ *        |
+ *        v
+ * rcl.shutdown(context)
+ *        or
+ * context.shutdown() ---> shutdown
+ * ```
+ * Must call rclnodejs.init(context) to initialize the context
+ * to the usable 'initialized' (valid) state be using.
```
minggangw, line 44: Thanks for adding this state transition diagram.

lib/context.js
```diff
@@ -19,64 +19,201 @@ const rclnodejs = require('bindings')('rclnodejs');
...
+   * @returns {undefined}
+   * @throws {Error} If there is a problem shutting down the context.
+   */
+  tryShutdown() {
+    if (this.isValid()) {
+      this.shutdown();
+    }
+  }
+
+  onNodeCreate(node) {
```
minggangw: nit: how about `onNodeCreated`?

lib/context.js
```diff
@@ -19,64 +19,201 @@ const rclnodejs = require('bindings')('rclnodejs');
...
+
+    if (this._nodes.includes(node)) {
+      // do nothing
+      return;
+    }
+
+    this._nodes.push(node);
+  }
+
+  onNodeDestroy(node) {
```
minggangw: nit: how about `onNodeDestroyed`?

minggangw (approved):
Thanks for your PR, good practice of OOP and we should have had it earlier. LGTM!

wayneparrott:
@minggangw Thx for the feedback. 
I wanted to get early positive feedback on the general idea of this PR before doing the extra work of expanding our test suite to exercise the full api changes. A big TODO is for the PR to include updates the test suite to cover the Node and Context api changes before we consider the PR ready. (Perhaps I should have created this PR in draft status.)

Also I want to focus some attention and get feedback on the tagging of some rcl functions @deprecated such as createNode(), createLifecycleNode(), spin(), that are now public methods on Node. I'm not opposed to removing the @deprecated tag if we feel OK with supporting the current functional api approach and this newer object-oriented api approach going forward. Thoughts?

minggangw:
It seems that the [Last successful build (#2125)](https://ci.ros2.org/view/packaging/job/packaging_linux/) has some problem and the ci fails constantly, I am going to re-trigger it once a new packaging is generated.

minggangw:
Would you please take a look at the error reported by [lint](https://travis-ci.org/github/RobotWebTools/rclnodejs/builds/750445596#L25279)

```bash
/root/rclnodejs/types/node.d.ts
  121:1  error  Replace `⏎` with `··`  prettier/prettier
  123:1  error  Insert `··`            prettier/prettier
  124:4  error  Insert `··`            prettier/prettier
  125:1  error  Insert `··`            prettier/prettier
  126:4  error  Insert `··`            prettier/prettier
  127:1  error  Insert `··`            prettier/prettier
  128:4  error  Insert `··`            prettier/prettier
  129:1  error  Insert `··`            prettier/prettier
```

minggangw:
Merged, thanks for your effort!

PR #772, opened 2021-01-19
Use null-terminated string in the tests when sending a raw topic

### LINKED ISSUE #771: Append '\0' when pusblishing a raw string by Buffer
Recently, the `test-raw-pub-sub.js` began to fain consistently. 

https://github.com/RobotWebTools/rclnodejs/blob/31efa7c28e8d20bcd6fe1311fb10e9afa7bd63a7/test/test-raw-pub-sub.js#L42-L51

```bash
  1) rclnodejs publisher test suite
       Publish serialized messages:

      Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal: 1 !== 0

      + expected - actual

      -1
      +0
```

The root cause is that the received string is terminated by a null character `\0`, so we should add the `\0` when sending the raw topic accordingly.

minggangw:
Fix #771

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -602,7 +602,7 @@ class Node extends rclnodejs.ShadowNode {
    * @param {boolean} options.enableTypedArray - The topic will use TypedArray if necessary, default: true.
    * @param {QoS} options.qos - ROS Middleware "quality of service" settings for the subscription, default: QoS.profileDefault.
    * @param {boolean} options.isRaw - The topic is serialized when true, default: false.
-   * @param {SubscriptionCallback} callback - The callback to be call when receiving the topic subscribed.
+   * @param {SubscriptionCallback} callback - The callback to be call when receiving the topic subscribed. The topic will be an instance of null-terminated Buffer when options.isRaw is true.
    * @return {Subscription} - An instance of Subscription.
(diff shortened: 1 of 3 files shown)


wayneparrott:
@minggangw I think it would be helpful to extend the comment on Publisher#publish() to note the null terminated string requirement. If you update the doc please update the publisher.d.ts api accordingly.
https://github.com/RobotWebTools/rclnodejs/blob/31efa7c28e8d20bcd6fe1311fb10e9afa7bd63a7/lib/publisher.js#L41

Also  I notice the raw publisher example does not use null terminated string.
https://github.com/RobotWebTools/rclnodejs/blob/31efa7c28e8d20bcd6fe1311fb10e9afa7bd63a7/example/publisher-raw-message.js#L35

minggangw:
The description of #771 may mislead the readers. I'd like to clarify it:

- The JavaScript doesn't use a null-terminated string as C does, so it's not necessary to add it.
- The buffer received comes from C functions, I guess it adds `\0` recently which causes the failure. In order to pass the unit test, we have to use a null-terminated string when comparing.

I think we should update the comments for the subscription instead, what do you think?

wayneparrott (approved):
LGTM

PR #801, opened 2021-07-07
add .gitattributes to project

wayneparrott:
Presently when a developer pulls code on windows his editor may show a zillion prettier errors due to the Windows default EOL does not match the default prettier EOL of LF. By adding a .gitattributes we can enforce the EOL to be LF (compatible with prettier) on all platforms by default.

This .gitattributes file is based on this [example](https://github.com/alexkaratarakis/gitattributes/blob/master/Web.gitattributes). I added some additional EOL config for *.[json|yaml|cpp|hpp] and a few others.  Please review and propose any changes you feel better suit the project and developer experience.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,117 @@
+# Based on https://github.com/alexkaratarakis/gitattributes/blob/master/Web.gitattributes
+
+# Auto detect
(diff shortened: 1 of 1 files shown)


minggangw (approved):
LGTM!

PR #811, opened 2021-09-26
Update the int64-napi to 2.0.9

minggangw:


wayneparrott:
I noticed ver 2.x of int64-napi is a node addon that requires a Rust toolchain to install. https://www.npmjs.com/package/int64-napi

minggangw:
Yes, we cannot expect users to install the Rust toolchain. I think we can pin to this version for int64-napi or find another alternative. So let me close this PR.

PR #831, opened 2022-01-07
fix: action server types

JGAntunes:
**Public API Changes**

Changes for the TS types for:
- `ExecuteCallback` for the action server
- `GoalCallback` for the action server

**Description**

Both of these callbacks seem to have wrong types setup. The first one should return a `Promise` which is then resolved in:
- https://github.com/RobotWebTools/rclnodejs/blob/73f40365e2257c372db446274af63f7ac1e4f010/lib/action/server.js#L359-L361

As for the `GoalCallback` type as [per the docs](http://robotwebtools.org/rclnodejs/docs/0.20.0/ActionServer.html):
- "The purpose of the goal callback is to decide if a new goal should be accepted or rejected. **The callback should take the goal request message as a parameter and must return a GoalResponse value.** "

And based on the code:
- https://github.com/RobotWebTools/rclnodejs/blob/73f40365e2257c372db446274af63f7ac1e4f010/lib/action/server.js#L259-L261

I've tested out both of these changes in a local project and they work as expected 👍

### DIFF
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -517,9 +517,7 @@ actionServer.registerGoalCallback();
 actionServer.registerCancelCallback();
 
 // $ExpectType void
-actionServer.registerExecuteCallback(() =>
-  Promise.resolve(new Fibonacci.Result())
-);
+actionServer.registerExecuteCallback(() => new Fibonacci.Result());
 
 // $ExpectType void
 actionServer.destroy();
@@ -557,5 +555,5 @@ function executeCallback(
   // $ExpectType void
   goalHandle.succeed();
 
-  return Promise.resolve(new Fibonacci.Result());
+  return new Fibonacci.Result();
 }
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -517,7 +517,9 @@ actionServer.registerGoalCallback();
 actionServer.registerCancelCallback();
 
 // $ExpectType void
-actionServer.registerExecuteCallback(() => new Fibonacci.Result());
+actionServer.registerExecuteCallback(() =>
+  Promise.resolve(new Fibonacci.Result())
+);
 
 // $ExpectType void
 actionServer.destroy();
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -555,5 +555,5 @@ function executeCallback(
   // $ExpectType void
   goalHandle.succeed();
 
-  return new Fibonacci.Result();
+  return Promise.resolve(new Fibonacci.Result());
 }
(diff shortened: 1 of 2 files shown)


wayneparrott:
@JGAntunes Thx for identifying the typing issue. 
dtslint is failing in test/types/main.ts due to the api changes and needs to be update:
```javascript
Error: C:/proj/rclnodejs/test/types/main.ts:507:3
ERROR: 507:3   expect  TypeScript@4.6 compile error: 
Argument of type '(goalHandle: ServerGoalHandle<"rclnodejs_test_msgs/action/Fibonacci">) => Fibonacci_Result' is not assignable to parameter of type 'ExecuteCallback<"rclnodejs_test_msgs/action/Fibonacci">'.
  Type 'Fibonacci_Result' is missing the following properties from type 'Promise<Fibonacci_Result>': then, catch, finally, [Symbol.toStringTag]
ERROR: 520:44  expect  TypeScript@4.6 compile error: 
Type 'Fibonacci_Result' is not assignable to type 'Promise<Fibonacci_Result>'.
    at C:\proj\rclnodejs\node_modules\dtslint\bin\index.js:207:19
...
```
Please update main.ts and we can wrap up this PR.

wayneparrott (dismissed):

wayneparrott:
@JGAntunes  would you also review test/test-action-server.js as I believe there is some inconsistency with executioncallback.

fampinheiro:
@wayneparrott I reverted the type change. I expanded it to accept a callback,  the callback returns the previous type or a promise that resolves for that type.

Let me know if you are okay with this change.

wayneparrott:
@JGAntunes  @fampinheiro appreciate you identifying the callback return type inconsistencies. Additionally in server.js I noticed that cancelCallback is awaited on yet the defaltCancelCallback fn does not return a promise. This leads me to believe that the original coder is attempting to account for the case where the executionCallback and cancelCallback functions could be long running async operations. Thus I believe cancelCallback definition should be modified similarly to your proposed modification of the exeutionCallback return type of `Promise<ActionResult<T>> | ActionResult<T>`.  Something like this:
```
type CancelCallback = () => CancelResponse | Promise<CancelResponse>;
```
Let's loop @minggangw and others that may know the action api better than me for their thoughts.

minggangw:
> Thus I believe cancelCallback definition should be modified similarly to your proposed modification of the exeutionCallback return type of Promise<ActionResult<T>> | ActionResult<T>.

I agree, we need to change this accordingly, and it can be separated into another PR, thanks!

minggangw (approved):
Thanks for your PR, @wayneparrott pls help to merge it if we decide to submit another PR to resolve the comments.

wayneparrott:
@JGAntunes again thx for identifying this issue and the suggested fix. The fix LGTM. Are you able to submit a similar fix for the cancelCallback? If no, I'll open a separate PR.

JGAntunes:
Hey @wayneparrott 👋 I can look into it today 👍

wayneparrott:
@JGAntunes no problem. I decided to move forward and merge this PR. I'll open a 2nd PR for the cancelCallback change.

PR #833, opened 2022-01-12
fix: ActionServer CancelCallback type definition

wayneparrott:
The internal actionserver code treats the cancelcallback as an awaitable
object. Yet the CancelCallback type definition specifies only a return
type of CancelResponse. The proper CancelCallback function definition
should return a union of CallbackResponse or a Promise that resolves to
a CallbackResponse.

This fix is related to fix #831 contributed by @JGAntunes


minggangw (approved):
LGTM, thanks!

PR #843, opened 2022-04-02
Free memory allocated for typed arrays

### LINKED ISSUE #842: Memory Leak when subscribing to messages containing UInt8Array
**Description**
It seems memory is not getting released for subscribers subscribing to messages containing a UInt8Array.  If a sufficiently large array is published, the memory of the Node.js process can grow very quickly.

- Library Version: ^0.21.0
- ROS Version: foxy
- Platform / OS: Ubuntu 20.04.3 LTS (64-bit)

**Steps To Reproduce**
Attached is a small nodejs example of the issue I'm seeing.  In the example, I have set up a publisher to publish a large image every 100ms.  I've set up a subscriber to subscribe to that topic, and print `process.memoryUsage().rss` each time it receives a message.

The example can be run by running `npx ts-node server.ts` in a terminal window and `npx ts-node client.ts` in another window.

When both are running, you can see that the memory footprint of the node process running the client grows whenever a message is received.

**Expected Behavior**
Memory should not grow with each message

**Actual Behavior**
Memory grows with each message
[rosleak.zip](https://github.com/RobotWebTools/rclnodejs/files/8336169/rosleak.zip)

minggangw:
This patch fixed the memory leak when using typed arrays on nodejs > 10

Fix: #842

### DIFF
--- a/src/rcl_bindings.cpp
+++ b/src/rcl_bindings.cpp
@@ -1330,6 +1330,7 @@ NAN_METHOD(CreateArrayBufferFromAddress) {
   memcpy(backing->Data(), addr, length);
   auto array_buffer =
       v8::ArrayBuffer::New(v8::Isolate::GetCurrent(), std::move(backing));
+  free(addr);
 #endif
 
   info.GetReturnValue().Set(array_buffer);


wayneparrott:
@minggangw bummer that dtslint is failing the test suite because of transposed union types, e.g.,
```
Error: /home/runner/work/rclnodejs/rclnodejs/test/types/main.ts:319:1
ERROR: 319:1  expect  TypeScript@4.7 expected type to be:
  Duration | Time
got:
  Time | Duration
ERROR: 322:1  expect  TypeScript@4.7 expected type to be:
  Duration | Time
got:
  Time | Duration
```
We saw this a while back. 
I'll give it a look asap.

wayneparrott:
See PRs #847 and #848 which combined should fix the linux workflow failures.

PR #847, opened 2022-04-27
Accomodate TS 4.7's new union order in dtslint tests

wayneparrott:
Similar to our test failures, others have struggled with dtslint failing
(false negative) due to inconsistent union type ordering. It appears
that $ExpectType <union> can be extended to list different orderings for
union types (permutations). In our case we have encountered issues with
this rule:
```
// $ExpectType Duration | Time
```
Extending this rule with an alternative ordering:
```
// $ExpectType Duration | Time || Time | Duration
```

### DIFF
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -315,10 +315,10 @@ time1.secondsAndNanoseconds;
 // $ExpectType Time
 time1.add(duration1);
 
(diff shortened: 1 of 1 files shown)


minggangw (approved):
LGTM, thanks!

PR #848, opened 2022-04-27
Fixes linux-galactic workflow to load the proper test-msgs version

wayneparrott:
The linux-build-and-test-galactic.yml workflow loading the wrong test_msgs (ros-rolling-test-msgs). This change fixes the workflow to load ros-galactic-test-msgs.


minggangw (approved):
LGTM, thanks a lot!

PR #855, opened 2022-05-30
Fix issues when using parameters of type array

ejalaa12:
**Public API Changes**
None

**Description**

This PR fixes the following issues with parameters when the parameter type is an array. (in order of commits)

- the parameter type `PARAMETER_STRING_ARRAY` was not handled everywhere
- values of parameters of type `PARAMETER_[BYTE/INTEGER/DOUBLE]_ARRAY` are of type `TypedArray` which throws a`TypeError('incompatible value')` in the `validate()`  method. This is because `_validArray()` (more specifically `Array.isArray()` returns `false` when given a `TypedArray`
- `arrayElementType` of `PARAMETER_BYTE_ARRAY` was set to `PARAMETER_BYTE` which is a type that does not exist (`undefined`). We can use `PARAMETER_INTEGER` instead.

Unit tests were added to reproduce the bugs, then the next commits fix those bugs.

### DIFF
--- a/lib/parameter.js
+++ b/lib/parameter.js
@@ -89,25 +89,25 @@ class Parameter {
         value = parameterMsg.value.bool_array_value;
         break;
       case ParameterType.PARAMETER_BYTE_ARRAY:
-        value = parameterMsg.value.byte_array_value;
+        value = Array.from(parameterMsg.value.byte_array_value);
         break;
       case ParameterType.PARAMETER_DOUBLE:
         value = parameterMsg.value.double_value;
         break;
       case ParameterType.PARAMETER_DOUBLE_ARRAY:
-        value = parameterMsg.value.double_array_value;
+        value = Array.from(parameterMsg.value.double_array_value);
         break;
       case ParameterType.PARAMETER_INTEGER:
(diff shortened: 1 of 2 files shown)


minggangw (approved):
Thanks for fixing this and the unit test! yeah, we should convert it back to Array, @wayneparrott would you please take a look from your side and see if we need to change for TS accordingly, if not please help to merge it, many thanks!

ros2jsguy:
@minggangw @ejalaa12 will give it a look later today.

wayneparrott (approved):

wayneparrott (approved):

wayneparrott:
@ejalaa12 thx for helping improve parameter support. Your efforts have been most appreciated.

OT - may I ask how you are using rclnodejs? Also please open an issue if you have suggestions for improving the use of rclnodejs including the onboarding experience.

ejalaa12:
@wayneparrott you're welcome. It's a great job you guys are doing already.

We're using rclnodejs as a backend to the HMI that controls our ros2 robot and some other products.

We'll be sure to drop new PR/Issue when needed :)

Thanks for the merge !

Issue #858, opened 2022-06-02
Client only saves one callback at a time 


ejalaa12:
**Description**

When using a Service Client, and sending two request in a row, because the Client saves one callback at a time, only the latest is used when processing the response message from the service. (look at line 63)
https://github.com/RobotWebTools/rclnodejs/blob/2e05913fb0e818f0c3a58904764ef752902a9a5e/lib/client.js#L51-L64

- Library Version: latest develop
- ROS Version: galactic
- Platform / OS: Linux Ubuntu 20.04

**Steps To Reproduce**

Here is a minimal example that reproduces the bug (based on the `example/client-example.js`):

```js

'use strict';

const rclnodejs = require('../index.js');

rclnodejs
  .init()
  .then(() => {
    const node = rclnodejs.createNode('client_example_node');

    const client = node.createClient(
      'example_interfaces/srv/AddTwoInts',
      'add_two_ints'
    );

    // First call
    client.waitForService(1000).then((result) => {
      const request = {
        a: Math.floor(1),
        b: Math.floor(2),
      };
      if (!result) {
        console.log('Error: service not available');
        // rclnodejs.shutdown();
        return;
      }
      console.log(`Sending: ${typeof request}`, request);
      client.sendRequest(request, (response) => {
        console.log(`Result1: ${typeof response}`, response);
        // rclnodejs.shutdown();
      });
    });

    // Second call
    client.waitForService(1000).then((result) => {

      const request = {
        a: Math.floor(10),
        b: Math.floor(20),
      };
      if (!result) {
        console.log('Error: service not available');
        // rclnodejs.shutdown();
        return;
      }
      console.log(`Sending: ${typeof request}`, request);
      client.sendRequest(request, (response) => {
        console.log(`Result2: ${typeof response}`, response);
        // rclnodejs.shutdown();
      });
    });

    rclnodejs.spin(node);
    setTimeout(() => {
      rclnodejs.shutdown();
    }, 1000);
  })
  .catch((e) => {
    console.log(`Error: ${e}`);
  });
```

**Expected Behavior**

The console log output should be:

```
Sending: object { a: 1, b: 2 }
Sending: object { a: 10, b: 20 }
Result1: object { sum: 3 }
Result2: object { sum: 30 }

```

**Actual Behavior**

`Result2` is printed twice:

```
Sending: object { a: 1, b: 2 }
Sending: object { a: 10, b: 20 }
Result2: object { sum: 3 }
[...]

PR #859, opened 2022-06-03
Fix issue 858

ejalaa12:
**Public API Changes**

None

**Description**

> When using a Service Client, and sending two request in a row, because the Client saves one callback at a time, only the latest is used when processing the response message from the service.

fix issue https://github.com/RobotWebTools/rclnodejs/issues/858

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -28,7 +28,7 @@ class Client extends Entity {
     super(handle, typeClass, options);
     this._nodeHandle = nodeHandle;
     this._serviceName = serviceName;
-    this._sequenceNumber = 0;
+    this._pending_requests = new Map();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -28,7 +28,7 @@ class Client extends Entity {
     super(handle, typeClass, options);
     this._nodeHandle = nodeHandle;
     this._serviceName = serviceName;
-    this._sequenceNumber = 0;
+    this._pending_requests = new Map();
```
minggangw: I would suggest renaming it to `this._senquenceNumberToCallbackMap` to align with the JS naming convention

lib/node.js
```diff
@@ -204,13 +204,12 @@ class Node extends rclnodejs.ShadowNode {
       this._runWithMessageType(
         client.typeClass.Response,
         (message, deserialize) => {
-          let success = rclnodejs.rclTakeResponse(
+          let sequence_number = rclnodejs.rclTakeResponse(
```
minggangw: Please rename it to `sequenceNumber`

src/rcl_bindings.cpp
```diff
@@ -831,24 +831,21 @@ NAN_METHOD(RclTakeResponse) {
...
-      reinterpret_cast<rmw_request_id_t*>(malloc(sizeof(rmw_request_id_t)));
-  header->sequence_number = sequence_number;
+  rmw_service_info_t header;
 
   void* taken_response =
-      node::Buffer::Data(Nan::To<v8::Object>(info[2]).ToLocalChecked());
-  rcl_ret_t ret = rcl_take_response(client, header, taken_response);
-  free(header);
+      node::Buffer::Data(Nan::To<v8::Object>(info[1]).ToLocalChecked());
+  rcl_ret_t ret = rcl_take_response_with_info(client, &header, taken_response);
```
minggangw, line 839: `rcl_take_response_with_info()` was not there the time when we implemented this JS function binding, thanks for updating it! I believe some other rcl functions are deprecated by the cpp/py clients already, which we need to pay attention to.

minggangw (commented):
Thanks for fixing this quickly, some comments are below.

lib/client.js
ejalaa12: Thanks for pointing this out ! I just pushed a commit with the change

minggangw (approved):
Thanks for fixing this issue! lgtm and merging. Your contributions are always welcome :)

PR #860, opened 2022-06-08
rm rclnodejs_test_msgs, compile script, cpp & py dirs, consolidates linux workflows [#856]

### LINKED ISSUE #856: Remove cases that depends on test/rclnodejs_test_msgs under \test
As these tests become unstable and cause the CI failures frequently, we decide to remove the C++ project used to verify some interactions with other ROS2 clients.

wayneparrott:
The main objective of this PR is to remove the troublesome compile_scripts.js, rclnodejs_test_msgs pkg and related tests and cpp/py example clients from the repo.

1. removed test/rclnodejs_test_msgs and related tests
- removed scripts/compile_tests.js and package.json script references
- replaced reference to Fibonacci action msgs in test-action-*.js and ts
types test files with equivalent msgs in test_msgs package

2 - discontinued testing compatibility with cpp and py rcl ros clients
- removed test/cpp/ folder
- removed test/py/ folder

3 - Replace test-non-primitive-static-array.js with
test-non-primitive-array.js

4 - Updated linux workflows
- merged individual linux workflows into a single workflow for rolling, humble, galactic and foxy
- removed obsolete distro specific workflows

5 - updated npmjs-readme.md

6 - fixed rolling specific issues
bindings.gyp
test-raw-pub-sub.js

Fix #856


.github/workflows/linux-build-and-test.yml
```diff
@@ -0,0 +1,76 @@
...
+  workflow_dispatch:
+
+jobs:
+  build-humble-and-rolling: # humble and rolling distros
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [10.X, 12.x, 14.X, 16.X, 17.X]
+        ros_distribution:
```
minggangw: I have a question here, if we are submitting a PR targeting on `develop` branch and we only want to run the verification on the `rolling`, because sometimes the rolling may introduce some interface changes that will break the compiling on other releases. For the current configuration in this PR, it seems that all ROS2 distro will be selected?

Another case is when releasing a new version, we just want to run a specific branch on the corresponding ROS2 distro.

wayneparrott: Ah yes, this version of the linux-build-and-test.yml workflow is triggered by any pull_request or push events and runs build and test suite tasks for foxy, galactic, humble and rolling distros on node 10-17 (N*M=20 scenarios). Since we have been keeping rclnodejs backward compatible to foxy I went with this big workflow to identify compatibility issues back to foxy.

Proposal:
1 -  keep this workflow but convert it to only run manually. This will serve an all-in-one compatibility test workflow that we launch manually thru the github actions UI.
2 - impl a new workflow that is a 1-1 between branch and ros distro and triggered by pull_request and push events on the specific branch, e.g.,
| **BRANCH** | **ROS DISTRO** |
|:---:|:---:|
| develop | rolling |
| humble-hawksbill | humble |
| galactic-geochelone | galactic |
| foxy-fitzroy | foxy |

This new workflow would for example run the galactic-geochelone branch against only the galactic ros distro for PR and push events on the branch.

Thoughts?

OT - I have a windows version of the workflow running reliably on a branch that I'll revise to match our linux solution once complete.

minggangw: The proposal is reasonable, we can make `linux-build-and-test.yml` run manually, so we can check the compatibility on different ROS distros, I believe many developers care about it. (it seems that our latest `develop` branch can run from Foxy to Rolling, which is great 👍 ).

The ideal case is that we can always make it compatible with as many ROS releases as possible, but I'm afraid this is very difficult and I don't want we spend much time solving this kind of issue or it blocks our daily work due to the failure of CI. So please go ahead as you proposed, thanks!

After the windows action is ready, I think we can clean up the table for the build status in the README file :)

wayneparrott: @minggangw [f1fd60a](https://github.com/RobotWebTools/rclnodejs/pull/860/commits/f1fd60a8280253a84e937ad733301e9e08d01c85) commit should address the requirements (I hope). 

I separated the linux and window ci into their own workflows. 

The following workflows are triggered by PR and push events on any of the foxy, galactic, humble or develop branches. They run ci for the changed branch against it's associated ros distro, i.e., foxy-fitzroy branch -> foxy distro.
windows-build-and-test.yml
linux-build-and-test.yml 

The following workflows are manually triggered and run a comprehensive ci of the `develop` branch against the foxy, galactic, humble and rolling distros:
windows-build-and-test-compatibility.yml
linux-build-and-test-compatibility.yml 

I was not able to fully tests these workflows in my dev env. (fingers crossed).

Also I doubt the workflows will work properly against the foxy and galactic branches without backporting the changes of this PR to those branches.

wayneparrott:
Odd that only the windows CI is failing the `test-raw-pub-sub.js` test on the rolling . I would have expected linux CI to fail in similar manner. 
```
1) rclnodejs publisher test suite
       Publish raw serialized messages:
      Uncaught AssertionError [ERR_ASSERTION]: distro: rolling, buffer: Hello ROS World, msg: Hello ROS World
      + expected - actual
      -1
      +0
      at Subscription._callback (test\test-raw-pub-sub.js:58:16)
...
```

wayneparrott:
poop! the most recent change to test-raw-pub-sub just move the fail from windows to linux os. Will investigate this more but does this seem legit for there to be difference in msg content between win and linux os'es?

minggangw:
I think we can disable `test-raw-pub-sub.js` temporarily on rolling if the failure is caused by the null-terminated issue because the ROS2 release changes back and forth and it seems that the behavior is different on Linux and Windows now. We don't have to change our tests accordingly, as the rolling itself is not a stable release. I don't want we spend so much time in solving kind of problem, what do you think?

wayneparrott:
Agree with disabling the test for now. Will update shortly. Thx

minggangw (approved):
Thanks for making our CI back online for both Linux & Windows, great! LGTM!

PR #867, opened 2022-07-22
Fix ActionServer cancelCallback declaration, doc & test

### LINKED ISSUE #866: ActionServer cancelRequestHandler() missing goalHandler param
While implementing an ActionServer I noticed the CancelCallback fn declaration is missing the GoalHandle parameter.

https://github.com/RobotWebTools/rclnodejs/blob/755a7a66f0b6bbd093a1f8d8cfcdefd3f64ea5f8/types/action_server.d.ts#L87

Based on the [ROS2 C++ action api](https://docs.ros2.org/foxy/api/rclcpp_action/classrclcpp__action_1_1Server.html#a1247929a1f116711b2056a7a254cfb51)  the CancelCallback declaration should be:
```typescript
type CancelCallback<T extends TypeClass<ActionTypeClassName>> = (
    goalHandle: ServerGoalHandle<T>
  ) => Promise<CancelResponse> | CancelResponse;
```

Additionally, test-action-server.js includes 2 test cases that need small modification to reflect this change.

wayneparrott:
While the ActionServer implementation will pass it's CancelCallback function a ServerGoalHandler argument, this fact is not clearly documented, not illustrated in the actionserver tests and the TS CancelCallback function declaration does not include a ServerGoalHandle parameter.
 
This PR updates docs, test suite and modifies the TypeScript CancelCallback function declaration to include a required ServerGoalHandle parameter. The change breaks existing TypeScript ActionServers that implement a custom CancelCallback function with a compilation error.  

Question: To make this a non-breaking change we can make the ServerGoalHandle optional. Thoughts?

Fix #866

### DIFF
--- a/types/action_server.d.ts
+++ b/types/action_server.d.ts
@@ -85,7 +85,7 @@ declare module 'rclnodejs' {
     goalHandle: ServerGoalHandle<T>
   ) => void;
   type CancelCallback<T extends TypeClass<ActionTypeClassName>> = (
-    goalHandle: ServerGoalHandle<T>
+    goalHandle?: ServerGoalHandle<T>
   ) => Promise<CancelResponse> | CancelResponse;
 
   interface ActionServerOptions extends Options<ActionQoS> {
--- a/types/action_server.d.ts
+++ b/types/action_server.d.ts
@@ -84,7 +84,9 @@ declare module 'rclnodejs' {
   type HandleAcceptedCallback<T extends TypeClass<ActionTypeClassName>> = (
     goalHandle: ServerGoalHandle<T>
   ) => void;
-  type CancelCallback = () => Promise<CancelResponse> | CancelResponse;
+  type CancelCallback<T extends TypeClass<ActionTypeClassName>> = (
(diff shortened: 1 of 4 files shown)


types/action_server.d.ts
```diff
@@ -153,14 +155,14 @@ declare module 'rclnodejs' {
...
-     * The callback should take one parameter containing the cancel request and must return a
+     * The callback should take one parameter containing the cancel request ( a GoalHandle) and must return a
      * {@link CancelResponse} value.
      *
      * There can only be one cancel callback per {@link ActionServer}, therefore calling this
      * function will replace any previously registered callback.
      * @param cancelCallback - Callback function, if not provided, then unregisters any previously registered callback.
      */
-    registerCancelCallback(cancelCallback?: CancelCallback): void;
+    registerCancelCallback(cancelCallback?: CancelCallback<T>): void;
```
minggangw, line 165: Optional param makes sense, thanks!

minggangw (approved):
Thanks for updating the doc & examples, LGTM!

PR #869, opened 2022-07-22
Add support for selective object import

### LINKED ISSUE #868: Unable to selectively import from rclnodejs
Recently I impl'ed a small utility in TypeScript that only required a few objects from rclnodejs. So I used this import syntax:

```typescript
import { init, Node } from 'rclnodejs';

async function main() {
  await init();
  const node = new Node('foo');
  node.spin();
}
```

This example fails with the following stacktrace:
```
    if (this._rosVersionChecked) {
             ^

TypeError: Cannot read properties of undefined (reading '_rosVersionChecked')
    at init (/home/wayne/dev/ros_utils/node_modules/rclnodejs/index.js:248:14)
...
```

Because of the stateful nature of the rcl object of index.js it is not possible to selectively import the init fn and by extension any objects.

This failure is caused by the `init()` function's `this` not being bound to the stateful rcl object of index.js. A quick fix is to move the `_rosVersionChecked` property to the global scope of index.js. This small change enables fine-grained import, e.g., `import { init, Node} from rclnodejs` rather than the classic wildcard import from rclnodejs, e.g., `import * as rclnodejs from 'rclnodejs'` 

I hope this makes sense. I'll submit a PR for consideration that should make this more clear.

wayneparrott:
This PR moves _rosVersionChecked to a module-level scope. By doing this the init() function can access it without it being bound to `this`.
Fix #868

### DIFF
--- a/index.js
+++ b/index.js
@@ -80,13 +80,13 @@ async function getCurrentGeneratorVersion() {
   });
 }
 
+let _rosVersionChecked = false;
+
 /**
  * A module that exposes the rclnodejs interfaces.
  * @exports rclnodejs
  */
 let rcl = {
-  _rosVersionChecked: false,
-
   /** {@link Clock} class */
   Clock: Clock,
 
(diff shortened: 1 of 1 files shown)


minggangw (approved):
LGTM, thanks!

PR #881, opened 2022-11-14
Update node.js

xhy279:
use let instead of const to fix `Cannot assign to "descriptors" because it is a constant` during bundling


minggangw (approved):
lgtm, thanks for finding and fixing this issue!

PR #891, opened 2023-01-31
Custom filter for omitting ROS2 interfaces from the message-generation process

### LINKED ISSUE #890: message generation fails on Rolling due to unsupported *.idl files in ros2bag pkg
The Rolling distro for windows includes the rosbag2_storage_mcap_testdata pkg. In this package there is a the  ComplexMsgDependsOnIdl.msg which includes BasicIdl.idl. The rclnodejs message generator only parses *.msg and *.srv files at this time. Thus the message generator fails with an error. Note that message generation is part of the npm install process for rclnodejs. Thus rclnodejs 0.21.4 can not be used with the Rolling distro on Windows at this time.

wayneparrott:
In response to the the blocking issue #890, I've introduced a simple mechanism for filtering ros2 interfaces out of the message generation process. The mechanism reads and external blocklist.json file containing filters that are used by `rosidl_gen/package.js` to determine which ROS2 interface files to excluded during the message-generation process. The user is informed of the ROS interfaces that are omitted with a `console.log` message. 

An ideal solution to #890 is to generate javascript message files from idl files. This will take some time during which we are blocked. This mechanism provides a quick work-around. And we can always disable this feature by removing the `rosidl_gen/blocklist.json` file in the future.
 
Also I switched the windows workflows to use node 18.12.0 in place of 18.13.0. There is a repeatable issue with node-gyp configuration on node 18.13. It seems to be related to the node cache. Switching to 18.12 avoids using a cached version of node 18 and the issue no longer occurs.

Fix #890

### DIFF
--- a/.github/workflows/windows-build-and-test-compatibility.yml
+++ b/.github/workflows/windows-build-and-test-compatibility.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.X, 12.X, 14.X, 16.11.X, 17.X, 18.X, 19.X]
+        node-version: [10.X, 12.X, 14.X, 16.11.X, 17.X, 18.12.X, 19.X]
         ros_distribution:
           # - foxy
           - galactic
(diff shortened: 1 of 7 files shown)


minggangw (approved):
Sorry for the big delay, LGTM! Shall we add the usage of `rosidl_gen/blocklist.json` into our document? considering there may be new cases in the future or we will be responsible to update this file?

Issue #894, opened 2023-02-09
error is not defined in the client class


Sata51:
The error method is not imported here

https://github.com/RobotWebTools/rclnodejs/blob/d5acae6253a4f0d504124b6e9523fbb9b0032521/lib/client.js#L73

wayneparrott:
Thx for identifying this issue.
Propose replacing `error()` with a simple `throw <error-msg>`. Do you mind submitting a PR?

PR #895, opened 2023-02-10
Fixes error undefined, throw error instead

Sata51:
**Public API Changes**
Throw an error instead of node crash

**Description**
Since the error method is undefined, prefer to throw a clean error

Related to #894


wayneparrott (approved):
LGTM
Thx!

PR #900, opened 2023-02-28
limit ci test matrix to only LTS and latest ROS & Nodejs distros

wayneparrott:
- updated actions: checkout - v3
setup-node - v3
setup-ros - v0.6

reduce node version matrix:
14, 16, 18, 19

reduce ros version matrix:
foxy, humble, rolling

On windows use explict nodejs version to workaround some silly github environment issue that is not correctly setting up node-gyp. I think this is caused by an environment/action caching of a bad instance.

Updated README.md to display the develop branch's compatibility with the LTS and current ROS branches on Linux and Windows.

**Public API Changes**

**Description**


wayneparrott:
Currently the linux ci is failing due to dependency on the system ubuntu 20 and 22 releases are not signed. This should resolve itself when the provider (microsoft) fixes the issue. Simply try running again may result in successful build.

wayneparrott:
self Approved

PR #901, opened 2023-03-02
Introducing content-filtering topics

wayneparrott:
ROS Humble introduced the content-filtering topics feature. This PR makes this feature available to rclnodejs developers.

**Resources**
[DDS 1.4 specification, Annex B](https://www.omg.org/spec/DDS/1.4/PDF)
[Content-filtering tutorial](https://docs.ros.org/en/rolling/Tutorials/Demos/Content-Filtering-Subscription.html)
https://github.com/ros2/design/pull/282
[subscription.h](https://github.com/ros2/rcl/blob/rolling/rcl/include/rcl/subscription.h)

**Supported RMWs**
Content-filtering is supported by the following RMWs as of the Humble release:
*  rmw_fastrtps - default Humble RMW
 * rmw_connext
 * rmw_gurumdds_cpp

Note-1: rmw_cyclonedds_cpp does NOT support content-filtering yet.
Note-2: Our experience is that content-filtering on rmw_fastrtps is only works on the Linux versions of ROS Humble and Rolling. Therefore, the test/blocklist.json excludes the test-subscription-content-filter.js test on Windows and Mac.   

**Key changes**
node.js
- added contentFilter to Options
- added static getDefaultOptions()
- updated createSubscription() to support contentFilter

node.d.ts
- added content-filter types

subscription.js
- isContentFilteringEnabled()
- setContentFilter()
- clearContentFilter()

subscription.d.ts
- updated with content-filter api

rcl_bindings.cpp
- added content-filtering to CreateSubscription()

rmw.js
- new class for identifying the current ROS middleware

test-subscription-content-filter.js
- test cases for content-filters

examples:
- publisher-content-filtering-example.js
- subscription-content-filtering-example.js

package.json
- added build/rebuild scripts for convenience


example/publisher-content-filter-example.js
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
```
minggangw: Please remove "Intel" branding

src/rcl_bindings.cpp
```diff
@@ -689,6 +747,83 @@ NAN_METHOD(CreateSubscription) {
...
+  char** argv = nullptr;
+
+  if (Nan::Has(contentFilter, Nan::New("parameters").ToLocalChecked())
+          .FromMaybe(false)) {
+    v8::Local<v8::Array> jsArgv = v8::Local<v8::Array>::Cast(
+        Nan::Get(contentFilter, Nan::New("parameters").ToLocalChecked())
+            .ToLocalChecked());
+    argc = jsArgv->Length();
+    if (argc > 0) {
+      argv = reinterpret_cast<char**>(malloc(argc * sizeof(char*)));
```
minggangw, line 805: You have to free these pointers explicitly, otherwise, it causes memory leak, please reference
https://github.com/RobotWebTools/rclnodejs/blob/feb8e03f5686ca48707ae13505f634b363f8e636/src/rcl_bindings.cpp#L93-L96

lib/subscription.js
```diff
@@ -66,6 +79,45 @@ class Subscription extends Entity {
...
+   *  the expression is undefined or an empty string.
+   * @param {object[]}  [contentFilter.parameters=undefined] - Array of objects that give values to
+   *  the ‘parameters’ (i.e., "%n" tokens) in the filter_expression. The number of supplied parameters must
+   *  fit with the requested values in the filter_expression (i.e., the number of %n tokens). default: undefined.
+   * @returns {boolean} - True if successful; false otherwise
+   * @see {@link https://www.omg.org/spec/DDS/1.4/PDF|DDS 1.4 specification, Annex B}
+   */
+  setContentFilter(contentFilter) {
+    if (!contentFilter) return false;
+
```
minggangw: Do we need to check `this.isContentFilteringEnabled()` here?

lib/node.js
```diff
@@ -484,6 +479,10 @@ class Node extends rclnodejs.ShadowNode {
       options = Object.assign(options, { isRaw: false });
     }
 
+    if (options.contentFilter === undefined) {
```
minggangw: I think we can omit this because `options.contentFilter` is already `undefined`.

src/rcl_bindings.cpp
```diff
@@ -652,19 +652,77 @@ NAN_METHOD(CreateSubscription) {
...
+            Nan::Get(contentFilter, Nan::New("parameters").ToLocalChecked())
+                .ToLocalChecked());
+        argc = jsArgv->Length();
+        if (argc > 0) {
+          argv = reinterpret_cast<char**>(malloc(argc * sizeof(char*)));
+          for (int i = 0; i < argc; i++) {
+            Nan::MaybeLocal<v8::Value> jsElement = Nan::Get(jsArgv, i);
+            Nan::Utf8String utf8_arg(jsElement.ToLocalChecked());
+            int len = utf8_arg.length() + 1;
+            argv[i] = reinterpret_cast<char*>(malloc(len * sizeof(char*)));
```
minggangw, line 708: Please free the pointers in `argv[i]` as comment below.

minggangw (commented):
Thanks for submitting this huge PR and implementing the content filter feature!

lib/subscription.js
wayneparrott: re: `isContentFilteringEnabled()`, this will only return true if both:
* the RMW implementation support content-filtering topics (CTF) and
* *the subscription is currently configured with a valid content-filter.

So the use of isContentFilteringEnabled() is limited to identifying if a subscription currently configured with a valid content-filter. I'm not aware of an rcl level api for identifying if an RMW implementation support CTFs.

I found this bit of low level [code in the fastrtps impl]( https://github.com/ros2/rmw_fastrtps/blob/1bf9e7a2d18caf6d50abf4aff1212b6d7dfc49f3/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L181) that implements `isContentFilteringEnabled()`. 

@minggangw do you think it would be more clear to developers if we renamed isContentFilteringEnabled() to something like:  `hasContentFilter()`?

 I originally implemented without the guard condition like this:
```javascript
  setContentFilter(contentFilter) {
    return contentFilter && contentFilter.expression
      ? rclnodejs.setContentFilter(this.handle, contentFilter)
      : this.clearContentFilter();
  }
```
At the rcl level, passing null or empty string for `conentFilter.expression` will clear the current content-filter. I propose that we use the implementation above. Thoughts?

minggangw: It sounds that `hasContentFilter()` is more clear.

>At the rcl level, passing null or empty string for conentFilter.expression will clear the current content-filter. I propose that we use the implementation above. Thoughts?

Agree, we can remove `if (!contentFilter) return false;` here, thanks

minggangw (approved):
LGTM! Thanks for making this big change!

PR #909, opened 2023-04-06
Fix CFT test suite fail

wayneparrott:
CI frequently fails due to what I believe are timing challenges between various timers and child-processes. This PR extends the timing intervals for the setContentFilter test-case in test-subscription-content-filter.js which I've seen fail periodically.


wayneparrott:
CI seems to run more reliably by lengthening the time of various timers in the CFT test suite.

minggangw:
Thanks!

PR #912, opened 2023-04-11
Fix raw msg processing bug

### LINKED ISSUE #911: possible bug when processing raw subscription message?
@minggangw 
In `node.js#execute()` should the following highlighted line be `continue` rather than `return`?

image

https://github.com/RobotWebTools/rclnodejs/blame/369e765f38fd0560876bc36e5dc5d85a2ac16171/lib/node.js#L188

wayneparrott:
This PR replaces a premature `return` statement in `node#execute()` with a `continue` statement when processing raw messages.

Note: CI build may fail on windows atm due to a change in rcl include file dependencies that will be updated in a separate PR #913.

Fix #911

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -185,7 +185,7 @@ class Node extends rclnodejs.ShadowNode {
         if (rawMessage) {
           subscription.processResponse(rawMessage);
         }
-        return;
+        continue;
       }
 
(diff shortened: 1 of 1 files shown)


minggangw (approved):
lgtm, thanks for fix it quickly!

PR #913, opened 2023-04-11
Fixes Windows build issues on rolling

wayneparrott:
* Adds rosidl_dynamic_typesupport to windows include path
* Exclude rosbag2_test_msgdefs from msg gen on windows. The problematic idl-only test msg in rosbag2_storage_mcap_testdata has been moved to rosbag2_test_msgdefs. Thus we will continue to exclude this package until msg generation directly accepts idl files in addition to msg/srv files.


minggangw (approved):
lgtm, thanks a  lot!

PR #915, opened 2023-04-14
refactored ctf test-subscription-content-filter to improve reliability

wayneparrott:
The `test-subscription-content-filter.j`s test-suite has proven to be unstable when running in CI, e.g., frequent timeout waiting for msgs published by external publisher processes. My initial response (hack) was to increase wait times. This improved reliability some but still there are frequent failures due to timeout. To improve the reliability of this test-suite I've rewritten it to run test msg publishers in a single process along with the CTF subscriptions. Publishers and subscriptions run on different nodes. This seems to be much better reliability in my CI test environment.


minggangw (approved):
Thanks for adding these well-designed cases, it ensures the quality! gltm

PR #920, opened 2023-05-30
[Iron] Add the iron release into distro.js

### LINKED ISSUE #919: Add distro information for ROS2 iron
We need to add `iron` into `distro.js`, making the compiling works properly.

minggangw:
This patch adds the iron release into distro.js and updates the unit test accordingly.

Fix: #919


minggangw:
@wayneparrott please take a look, and we should prepare another `rclnodejs` release for `iron` later.

PR #945, opened 2023-11-30
fix: node.createService with async callback

### LINKED ISSUE #944: Node.createService does not work with async callback
**Description**

It seems `Node.createService`  has some problems when dealing with  a asynchronous callback functions.

- Library Version: 0.23.2
- ROS Version: humble
- Platform / OS: Ubuntu 22.04.3 LTS (running in docker container)

**Steps To Reproduce**

``` js
const rclnodejs = require('rclnodejs');

function createNode() {
  const node = new rclnodejs.Node('bug');
  node.createService('rclnodejs_bug/srv/GetAnswer', 'bug/get_answer', async (req, rsp) => {
    const response = rsp.template;
    response.answer = 42;
    setTimeout(()=>rsp.send(response), 100);
  });
  return node;
};

(async ()=>{
  await rclnodejs.init();
  const node = createNode();
  node.spin();
})();
```

I run the above code using node (20.10.0).

Then on a second terminal, I run ``ros2 service call /bug/get_answer rclnodejs_bug/srv/GetAnswer``

**Expected Behavior**

```
requester: making request: rclnodejs_bug.srv.GetAnswer_Request()
^Chlr@kikuchiyo:/workspaces/pwd$ ros2 service call /bug/get_answer rclnodejs_bug/srv/GetAnswer 
requester: making request: rclnodejs_bug.srv.GetAnswer_Request()

response:
rclnodejs_bug.srv.GetAnswer_Response(answer=42)
``` 

**Actual Behavior**
```
/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/generated/rclnodejs_bug/rclnodejs_bug__srv__GetAnswer_Response.js:64
        throw new TypeError('Invalid argument: answer in GetAnswer_Response');
        ^

TypeError: Invalid argument: answer in GetAnswer_Response
    at GetAnswer_ResponseWrapper.freeze (/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/generated/rclnodejs_bug/rclnodejs_bug__srv__GetAnswer_Response.js:64:15)
    at GetAnswer_ResponseWrapper.serialize (/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/generated/rclnodejs_bug/rclnodejs_bug__srv__GetAnswer_Response.js:69:10)
    at Service.processRequest (/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/lib/service.js:82:44)
    at /workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/lib/node.js:235:21
    at Node._runWithMessageType (/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/lib/node.js:1654:5)
    at Node.execute (/workspaces/pwd/rclnodejs_bug/node_modules/rclnodejs/lib/node.js:226:12)

Node.js v20.10.0
```

**Workaround**

Remove the `async` keyword from the callback. If necessary, move code that uses `await` and friends to a separate function. 

**What I think is the problem**

Have a look at this piece of code in service.js:

https://github.com/RobotWebTools/rclnodejs/blob/4576609fa1941b2694eb1068eaccacf17cecf151/lib/service.js#L77-L84

It seems that the code expects callback to either return a response synchronously, or to call `response.send(...)`. The if condition in line 80 tries to determine if the latter has not happened yet and if in fact there is a return value. The problem here is that for an `async` function, there is *always* a return value: A Promise.

In my toy example above, this Promise object gets misinterpreted as an actual response value. Also, the check for `response.sent` does not help in this case, because the response is send asynchronous. So at the time that the if condition is evaluated, it has in fact not been sent *yet*.

The resulting behavior from the service module is that it will try to send a second response, ultimately failing (because the returned value is not a response, but a Promise resolving to undefined), and causing a considerable amount of confusion in for example a mocha test suite that was executing this.

As for possible mitigations, I am not sure. Maybe it would help to simply ignore the return value? I think this is undocumented behavior anyway?.

ldegen:
This should fix #944 

**Public API Changes**
This patch should enable `createService`  to correctly deal with `async` callback functions.
All previously supported "modes" should still work as before.

**Description**

The only thing that changes is that we now wrap any value returned by the callback in a `Promise.resolve(...)`.
We wait for this to resolve before continuing. For a callback that returns a response synchronously, this should
resolve right away (next tick, probably, depending on the implementation of Promise), so there should not be any noticable difference for those cases. If the callback returns nothing (or undefined), the same logic as before will prevent
us from trying to send the response twice. Only it now also works correctly if the return value is a Promise.

**Notes**

- I am not sure if I put my test case in the right spot.
- I am getting 153 linter errors for code that I never touched. They all look like this:
 ```
/workspaces/pwd/rclnodejs/test/utils.js
  0:0  error  Parsing error: require() of ES Module /workspaces/pwd/rclnodejs/node_modules/eslint/node_modules/eslint-scope/lib/definition.js from /workspaces/pwd/rclnodejs/node_modules/babel-eslint/lib/require-from-eslint.js not supported.
  Instead change the require of definition.js in /workspaces/pwd/rclnodejs/node_modules/babel-eslint/lib/require-from-eslint.js to a dynamic import() which is available in all CommonJS modules
  
  ```
  Maybe something wrong with the linter config?

### DIFF
--- a/lib/service.js
+++ b/lib/service.js
@@ -75,17 +75,18 @@ class Service extends Entity {
 
     const plainObj = request.toPlainObject(this.typedArrayEnabled);
     const response = new Response(this, headerHandle);
-    let responseToReturn = this._callback(plainObj, response);
+    Promise.resolve(this._callback(plainObj, response)).then((responseToReturn) => {
 
-    if (!response.sent && responseToReturn) {
-      responseToReturn = new this._typeClass.Response(responseToReturn);
-      const rawResponse = responseToReturn.serialize();
-      rclnodejs.sendResponse(this._handle, rawResponse, headerHandle);
-    }
+      if (!response.sent && responseToReturn) {
+        responseToReturn = new this._typeClass.Response(responseToReturn);
+        const rawResponse = responseToReturn.serialize();
+        rclnodejs.sendResponse(this._handle, rawResponse, headerHandle);
+      }
 
-    debug(
-      `Service has processed the ${this._serviceName} request and sent the response.`
-    );
+      debug(
+        `Service has processed the ${this._serviceName} request and sent the response.`
+      );
+    });
   }
 
   static createService(nodeHandle, serviceName, typeClass, options, callback) {
--- a/test/test-service-with-async-callback.js
+++ b/test/test-service-with-async-callback.js
@@ -1,0 +1,71 @@
+// Copyright (c) 2017 Intel Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const assert = require('assert');
+const rclnodejs = require('../index.js');
+
+describe('Test creating a service with an async callback', function(){
+
+  this.timeout(60 * 1000);
+
+  before(function () {
+    return rclnodejs.init();
+  });
+
+  after(function () {
+    rclnodejs.shutdown();
+  });
+
+  it('Node.createService() with async callback', function (done) {
+    // this should expose issue #944
+    var clientNode = rclnodejs.createNode('single_ps_client_async');
+    var serviceNode = rclnodejs.createNode('single_ps_service_async');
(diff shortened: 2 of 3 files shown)


minggangw (approved):
@ldegen Thanks for the PR, the fix makes sense, lgtm.

Per what you mentioned in the commit log, I prefer moving the test into a separate file before merging it, for example, `test-service-with-async-callback`, to make it clear.

ldegen:
> test-service-with-async-callback

Sure, I'll see to it.

ldegen:
> > test-service-with-async-callback
> 
> Sure, I'll see to it.

Done ([92877d2](https://github.com/RobotWebTools/rclnodejs/pull/945/commits/92877d2984deeb5d21dfee002590f2eaa8b86965))

minggangw:
@ldegen thanks for your PR, merged!

PR #954, opened 2024-01-25
Move rimraf to dependencies

chfritz:
**Public API Changes**
None

**Description**
`rimraf` is used in the `install` script (via rebuild -> clean) and as such is always required, not just in dev. `devDependencies` are not installed transitively, so when using rclnodejs in a package that is used in another project, rimraf is not installed, leading to an error during installation.


minggangw:
@chfritz Thanks for submitting the fix, approved!

minggangw (approved):

Issue #955, opened 2024-02-06
slam toolbox causes bugs


locmark:
**Description**

Using rclnodejs with slam-toolbox creates bug, that makes impossible to generate ros messages.

When installing rclnodejs this error does not occur but my custom ros messages are missing TypeScript definitions. Applying proposed patch fixes this problem. 

- Library Version: 0.26.0
- ROS Version: iron
- Platform / OS: ubuntu 22.04.3 LTS

**Steps To Reproduce**
```
sudo apt install ros-iron-slam-toolbox
```
and in new console:
```
npx generate-ros-messages
```

**Expected Behavior**
```
Start generation of ROS2 JavaScript messages...
Generation complete.
```

**Actual Behavior**

```
Start generation of ROS2 JavaScript messages...
Caught error: Error: Cannot find module '../../generated/slam_toolbox/slam_toolbox__srv__AddSubmap_Request.js'
Require stack:
- /home/k/backend/node_modules/rclnodejs/generated/slam_toolbox/slam_toolbox__srvs__AddSubmap_Event.js
- /home/k/backend/node_modules/rclnodejs/lib/interface_loader.js
- /home/k/backend/node_modules/rclnodejs/rostsd_gen/index.js
- /home/k/backend/node_modules/rclnodejs/scripts/generate_messages.js
```

**quick fix:**
use 'patch-package' package with this file 
rclnodejs+0.26.0.patch
```
diff --git a/node_modules/rclnodejs/rostsd_gen/index.js b/node_modules/rclnodejs/rostsd_gen/index.js
index dbb9b64..6aca532 100644
--- a/node_modules/rclnodejs/rostsd_gen/index.js
+++ b/node_modules/rclnodejs/rostsd_gen/index.js
@@ -64,6 +64,7 @@ function getPkgInfos(rootDir) {
 
     for (let filename of files) {
       const typeClass = fileName2Typeclass(filename);
+      if (typeClass.type !== 'srv' && typeClass.type !== 'action' && typeClass.type !== 'msg') continue;
       if (
         !typeClass.type ||
         pkgFilters.matchesAny({
```

PR #956, opened 2024-02-06
typeclass check

locmark:
**Public API Changes**
None

**Description**
Fixes described issue:
https://github.com/RobotWebTools/rclnodejs/issues/955

For now, it looks like this package does not support a custom folder scheme like the "srvs" folder in slam-toolbox. That's why i propose this temporary solution.

### DIFF
--- a/rostsd_gen/index.js
+++ b/rostsd_gen/index.js
@@ -64,6 +64,7 @@ function getPkgInfos(rootDir) {
 
     for (let filename of files) {
       const typeClass = fileName2Typeclass(filename);
+      if (typeClass.type !== 'srv' && typeClass.type !== 'action' && typeClass.type !== 'msg') continue;
       if (
         !typeClass.type ||
         pkgFilters.matchesAny({


rostsd_gen/index.js
```diff
@@ -64,6 +64,7 @@ function getPkgInfos(rootDir) {
 
     for (let filename of files) {
       const typeClass = fileName2Typeclass(filename);
+      if (typeClass.type !== 'srv' && typeClass.type !== 'action' && typeClass.type !== 'msg') continue;
```
minggangw, line 67: Hi @locmark thanks for submitting the PR, my first thought is this way does hide the exception, but not solve it. I'm going to take a look later, thanks!

locmark: Yeah, it doesn't solve it at all. But if solving it neatly is going to take like forever, I consider it better than not working module. 
Just to be clear, I'm not implying that it should be merged, but maybe some temporary solution should be made.

minggangw: No worries, I have submitted a PR to fix it, please try with latest https://www.npmjs.com/package/rclnodejs/v/0.26.1. Thanks for your investigation and  your contributions are always welcome!

Issue #1074, opened 2025-03-10
UUID package is required to compile


kissadam21:
Hey all,

The UUID package is also required for the rclnodejs to compile. Looks like without it my simple lifecycle publisher code won't compile.

```
[node-1] node:internal/modules/cjs/loader:1031
[node-1]   throw err;
[node-1]   ^
[node-1] 
[node-1] Error: Cannot find module 'uuid'
[node-1] Require stack:
[node-1] - /devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/node_modules/rclnodejs/lib/action/uuid.js
[node-1] - /devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/node_modules/rclnodejs/lib/action/client.js
[node-1] - /devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/node_modules/rclnodejs/index.js
[node-1] - /devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/dist/publisher.js
[node-1]     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1028:15)
[node-1]     at Function.Module._load (node:internal/modules/cjs/loader:873:27)
[node-1]     at Module.require (node:internal/modules/cjs/loader:1100:19)
[node-1]     at require (node:internal/modules/cjs/helpers:119:18)
[node-1]     at Object.<anonymous> (/devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/node_modules/rclnodejs/lib/action/uuid.js:18:24)
[node-1]     at Module._compile (node:internal/modules/cjs/loader:1198:14)
[node-1]     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
[node-1]     at Module.load (node:internal/modules/cjs/loader:1076:32)
[node-1]     at Function.Module._load (node:internal/modules/cjs/loader:911:12)
[node-1]     at Module.require (node:internal/modules/cjs/loader:1100:19) {
[node-1]   code: 'MODULE_NOT_FOUND',
[node-1]   requireStack: [
[node-1]     '/devenv/UserApplication/ros2-hello-world/install/hello_world_js/share/hello_world_js/node_modules/rclnodejs/lib/action/uuid.js',
[...]

PR #1075, opened 2025-03-10
Move uuid to dependencies

kissadam21:
Without uuid the build fails. Solves my reported issue: https://github.com/RobotWebTools/rclnodejs/issues/1074


minggangw:
LGTM! thanks for fixing it!

Issue #1307, opened 2025-10-22
JSON-Safe Message Serialization


mahmoud-ghalayini:
### **The Problem**

When you receive messages in rclnodejs, numeric arrays come back as `TypedArray` objects (Float64Array, Int32Array, etc.). The problem? **These don't serialize to JSON properly**, which is a huge pain for web applications.

This breaks pretty much every common web pattern:

```javascript
node.createSubscription('sensor_msgs/msg/LaserScan', '/scan', (msg) => {
  console.log(typeof msg.ranges); // Float64Array
  
  // None of these work as expected:
  JSON.stringify(msg);                     // Loses data or returns {}
  res.json(msg);                           // Express can't serialize it
  websocket.send(JSON.stringify(msg));     // Sends broken data
  logger.info(JSON.stringify(msg));        // Can't log it properly
});
```

Whether you're building a dashboard, REST API, WebSocket stream, or just trying to log messages - you hit this issue immediately.

### **What We're Doing Now**

Everyone ends up writing their own message normalizer that recursively converts TypedArrays to regular arrays. Here's what that looks like:

```javascript
class MessageNormalizer {
  normalizeMessage(msg) {
    if (msg instanceof Float64Array) return Array.from(msg);
    if (msg instanceof Int32Array) return Array.from(msg);
    if (msg instanceof ArrayBuffer) return Array.from(new Float64Array(msg));
    // ... check every TypedArray type
    // ... handle nested objects
    // ... handle nested arrays
    // ... ~150 lines of boilerplate
  }
}
```

### **Proposed Solution**

Let users opt-in to JSON-safe messages with a simple option:

```javascript
// Just add serializationMode option
node.createSubscription(
  'sensor_msgs/msg/LaserScan', 
  '/scan', 
  (msg) => {
    // msg.ranges is now a regular array!
    websocket.send(JSON.stringify(msg)); // Just works
  },
  { 
    serializationMode: 'json'
  }
);
```

**Three modes to choose from:**
- `'typed'` (default) - Current behavior, keeps TypedArrays for performance
- `'plain'` - Converts TypedArrays to regular arrays
- `'json'` - Fully JSON-safe (handles BigInt, special types, etc.)

**Also expose utility functions for one-off conversions:**

```javascript
const rclnodejs = require('rclnodejs');

// When you need to convert a message manually
const jsonSafe = rclnodejs.toJSONSafe(message);
const jsonString = rclnodejs.toJSONString(message);
```

[...]

PR #1308, opened 2025-10-22
Add JSON-Safe serialization modes

mahmoud-ghalayini:
## **Public API Changes**

**New subscription option:**
- `serializationMode: 'default' | 'plain' | 'json'` for `createSubscription()`

**New utility functions:**  
- `rclnodejs.toJSONSafe(obj)` - makes objects JSON-friendly
- `rclnodejs.toJSONString(obj)` - converts directly to JSON string

**Fully backward compatible** - existing code works unchanged.

## **Description**

This adds **message serialization modes** to fix a common headache: ROS sensor messages use `TypedArray`s which don't play nice with JSON.

**The Problem:**
If you try to `JSON.stringify()` a LaserScan message, you get `{}` instead of the actual data. This breaks web APIs, logging, and any workflow that needs JSON.

**The Solution:**
Three modes to handle messages differently:

- **`'default'`** Uses native rclnodejs behavior
- **`'plain'`**: Convert TypedArrays to regular arrays - JSON works!  
- **`'json'`**: Handle everything (TypedArrays, BigInt, NaN) - fully JSON-safe

[#1307](https://github.com/RobotWebTools/rclnodejs/issues/1307)

### DIFF
--- a/example/topics/subscriber/subscription-json-utilities-example.js
+++ b/example/topics/subscriber/subscription-json-utilities-example.js
@@ -1,0 +1,41 @@
+// Copyright (c) 2024 rclnodejs contributors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('../../../index.js');
+
+/**
+ * This example demonstrates the JSON utility functions for manual message conversion.
+ * These utilities are useful when you need to convert messages on-demand
+ * rather than using the serializationMode subscription option.
+ */
+async function main() {
+  await rclnodejs.init();
+  const node = new rclnodejs.Node('json_utilities_example_node');
+
+  node.createSubscription('sensor_msgs/msg/LaserScan', '/laser_scan', (msg) => {
+    // Convert using utility functions
+    const jsonSafe = rclnodejs.toJSONSafe(msg);
+    const jsonString = rclnodejs.toJSONString(msg);
+
+    console.log(
+      `Original: ${msg.ranges ? msg.ranges.constructor.name : 'undefined'}, JSON-safe: ${jsonSafe.ranges ? jsonSafe.ranges.constructor.name : 'undefined'}, JSON length: ${jsonString.length}`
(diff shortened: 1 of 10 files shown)


minggangw:
@mahmoudalghalayini thanks for submitting the PR with ut and examples, that's great! The overall looks good to me, I will be oof for the rest of the week and will take a closer look early next week. Meanwhile, please feel free to take what we disscussed in #1306, thanks!

lib/subscription.js
```diff
@@ -55,7 +57,13 @@ class Subscription extends Entity {
     if (this._isRaw) {
       this._callback(msg);
     } else {
-      this._callback(msg.toPlainObject(this.typedArrayEnabled));
+      let message = msg.toPlainObject(this.typedArrayEnabled);
+
+      if (this._serializationMode !== 'typed') {
```
minggangw: I have some concern about the naming `typed` here, considering when:
1. `this.typedArrayEnabled` is `false`
2. `this._serializationMode` is `typed`

The message may contain plain array even if the array can be [typed-array](https://github.com/RobotWebTools/rclnodejs/blob/c8f0ae66e6367516352ce576e4ebd1e2657e6914/rosidl_gen/templates/message-template.js#L57), which may cause confusing.

When `typedArrayEnabled` is `true`, `rclnodejs` won't allocate additional memory from C++ side, instead, the typed array is backed by a `Buffer` if possible, whose memory was allocated by `rcl`, but I don't have a proper name in my mind now.

example/topics/subscriber/subscription-json-utilities-example.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2024 rclnodejs contributors. All rights reserved.
```
minggangw: I prefer unifying the wording here, you can either:
1. Add your name, like https://github.com/RobotWebTools/rclnodejs/blob/develop/lib/parameter_service.js
2. Use a general name, like https://github.com/RobotWebTools/rclnodejs/blob/develop/lib/utils.js

It's up to you. nit: it's year of 2025 already :)

lib/node.js
```diff
@@ -666,6 +675,10 @@ class Node extends rclnodejs.ShadowNode {
    * @param {boolean} options.enableTypedArray - The topic will use TypedArray if necessary, default: true.
    * @param {QoS} options.qos - ROS Middleware "quality of service" settings for the subscription, default: QoS.profileDefault.
    * @param {boolean} options.isRaw - The topic is serialized when true, default: false.
+   * @param {string} [options.serializationMode='default'] - Controls message serialization format:
```
minggangw, line 678: `default` makes sence to combine with `enableTypedArray`

minggangw (approved):
lgtm, thanks for submitting the PR!

minggangw:
@mahmoud-ghalayini this change has landed onto latest [v1.6.0](https://www.npmjs.com/package/rclnodejs/v/1.6.0?activeTab=readme), thanks for your contribution!

Issue #1310, opened 2025-10-27
Promise-Based Service Calls (async/await support)


mahmoud-ghalayini:
### **The Problem**

Service calls currently only support callbacks, which doesn't fit well with modern JavaScript patterns. You can't use `async/await`, making error handling awkward and code harder to read.

**Current (callback-based):**
```javascript
client.sendRequest(request, (response) => {
  // Can't easily handle errors
  // Can't use try/catch
  // Hard to chain multiple calls
  // No timeout support
});
```

**What we want (async/await):**
```javascript
try {
  const response = await client.sendRequest(request);
  // Clean, readable code
  // Proper error handling
  // Easy to chain calls
} catch (error) {
  // Handle errors naturally
}
```

### **Current Workaround**

We have to manually wrap every service call in a Promise:

```javascript
function callServiceAsync(client, request) {
  return new Promise((resolve, reject) => {
    const timeout = setTimeout(() => reject(new Error('Timeout')), 5000);
    
    client.sendRequest(request, (response) => {
      clearTimeout(timeout);
      resolve(response);
    });
  });
}

// Then use it everywhere
const response = await callServiceAsync(client, request);
```

This adds boilerplate to every service call in every project.

### **Proposed Solution**

Add async/await support to service calls with optional timeout and cancellation:

```javascript
// Simple promise-based call
const response = await client.sendRequestAsync(request);

// With timeout
const response = await client.sendRequestAsync(request, { timeout: 5000 });

// With AbortController (modern cancellation pattern)
const controller = new AbortController();
const response = await client.sendRequestAsync(request, { 
  signal: controller.signal 
});

// Cancel if needed
controller.abort();
```

### **Real Usage**

**Before:**
```javascript
client.sendRequest(request1, (response1) => {
  client.sendRequest(request2, (response2) => {
    client.sendRequest(request3, (response3) => {
      // Callback hell
    });
  });
});
```

**After:**
```javascript
try {
  const response1 = await client.sendRequestAsync(request1);
  const response2 = await client.sendRequestAsync(request2);
  const response3 = await client.sendRequestAsync(request3);
  // Clean sequential calls
} catch (error) {
  console.error('Service call failed:', error.message);
}
```

### **Implementation Notes**

- Keep existing `sendRequest(request, callback)` for backward compatibility
- Add new `sendRequestAsync(request, options)` method
- Support timeout via options or AbortSignal
- Works with all existing service types

[...]

PR #1311, opened 2025-10-27
feat: add promise-based service calls

mahmoud-ghalayini:
[#1310](https://github.com/RobotWebTools/rclnodejs/issues/1310)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -33,7 +33,7 @@ class Client extends Entity {
   }
 
   /**
-   * This callback is called when a resopnse is sent back from service
+   * This callback is called when a response is sent back from service
    * @callback ResponseCallback
    * @param {Object} response - The response sent from the service
    * @see [Client.sendRequest]{@link Client#sendRequest}
@@ -43,7 +43,7 @@ class Client extends Entity {
    */
 
   /**
-   * Send the request and will be notified asynchronously if receiving the repsonse.
+   * Send the request and will be notified asynchronously if receiving the response.
    * @param {object} request - The request to be submitted.
    * @param {ResponseCallback} callback - Thc callback function for receiving the server response.
    * @return {undefined}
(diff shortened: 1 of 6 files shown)


lib/client.js
```diff
@@ -33,7 +33,7 @@ class Client extends Entity {
   }
 
   /**
-   * This callback is called when a resopnse is sent back from service
+   * This callback is called when a response is sent back from service
```
minggangw, line 36: thanks for correcting the typo

lib/client.js
```diff
@@ -65,6 +65,88 @@ class Client extends Entity {
     this._sequenceNumberToCallbackMap.set(sequenceNumber, callback);
   }
 
+  /**
+   * Send the request and return a Promise that resolves with the response.
+   * @param {object} request - The request to be submitted.
+   * @param {object} [options] - Optional parameters for the request.
+   * @param {number} [options.timeout] - Timeout in milliseconds for the request.
```
minggangw, line 72: is it possible to leverage the [`AbortSignal.timeout()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static) to avoid passing additional param?

minggangw (commented):
@mahmoudalghalayini thanks for submitting the PR, just a comment about the parameter used for the added function. BTW, please feel free to add your name into [CONTRIBUTORS](https://github.com/RobotWebTools/rclnodejs/blob/develop/CONTRIBUTORS.md)

lib/client.js
mahmoud-ghalayini: I think we need to pass the parameter anyway, we would want to give the user the ability to set the `timeout` of the request, but we can definitely change from using `setTimeout` to `AbortSignal.timeout`.

mahmoud-ghalayini: It will look something like this

```
let effectiveSignal = options.signal;

if (options.timeout !== undefined && options.timeout >= 0) {
  const timeoutSignal = AbortSignal.timeout(options.timeout);

  if (options.signal) {
    // Combine user signal and timeout signal
    effectiveSignal = AbortSignal.any([options.signal, timeoutSignal]);
  } else {
    // Only timeout signal
    effectiveSignal = timeoutSignal;
  }
}

minggangw: This looks good to me 👍

mahmoud-ghalayini: Do we have a schedule for releases or when do we usually release a new version?

minggangw: Currently, we don't have a fixed period to release a new verison, it's on demand. Usaully, I would follow:

1. Ensure the lateset release can work properly when a new ROS2/Node.js release comes out.
2. When a critical bug fixed.
3. New feature introduced.
4. Some one requested.

If you check the release history, approximately, there is a new relase at least every one or two months.

minggangw (approved):
LGTM, thanks for your PR!

minggangw:
@mahmoud-ghalayini I have approved, please add some description when merging, thanks!

mahmoud-ghalayini:
> @mahmoud-ghalayini I have approved, please add some description when merging, thanks!

Description is added, but I don't have merge right :smile:

minggangw:
> > @mahmoud-ghalayini I have approved, please add some description when merging, thanks!
> 
> Description is added, but I don't have merge right 😄

@mahmoud-ghalayini I have added you as collaborator, please check your email inbox to see any invitation received.

Issue #1343, opened 2025-12-05
add MessageIntrospector for message schema inspection


mahmoud-ghalayini:
### Problem

When working with ROS 2 messages programmatically, developers often need to understand message structure:

```javascript
// How do I know what fields a Twist message has?
// Currently requires internal knowledge or using loader.loadInterface:
const loader = require('rclnodejs/lib/interface_loader.js');
const TwistClass = loader.loadInterface('geometry_msgs/msg/Twist');
console.log(TwistClass.ROSMessageDef); // Complex internal structure
```

Common needs:
- Discover field names for a message type
- Get default values to understand expected structure
- Access schema information for documentation or UI generation
- Debug message structure without diving into generated code

### Solution

Add a `MessageIntrospector` class that provides a simple API for message schema inspection:

```javascript
const { MessageIntrospector } = require('rclnodejs');

const Twist = new MessageIntrospector('geometry_msgs/msg/Twist');

// Discover fields
console.log(Twist.fields);    // ['linear', 'angular']

// Get default structure as plain object
console.log(Twist.defaults);  
// { linear: { x: 0, y: 0, z: 0 }, angular: { x: 0, y: 0, z: 0 } }

// Access schema for field types
console.log(Twist.schema.msgName);  // 'Twist'

// Works with any message type
const JointState = new MessageIntrospector('sensor_msgs/msg/JointState');
console.log(JointState.fields);  // ['header', 'name', 'position', 'velocity', 'effort']
```

PR #1346, opened 2025-12-09
Add MessageIntrospector for message schema inspection

mahmoud-ghalayini:
**Public API Changes**

Added `MessageIntrospector` class to inspect ROS 2 message structure without using `loader.loadInterface` directly. Useful for debugging, documentation, and building dynamic UIs based on message structure.

**Constructor:**

- `new MessageIntrospector(typeName)` - Create an introspector for any message type

**Properties:**

- `introspector.typeName` - Get the full message type name
- `introspector.typeClass` - Get the underlying message class constructor
- `introspector.fields` - Get array of field names
- `introspector.schema` - Get ROSMessageDef schema object
- `introspector.defaults` - Get default values for all fields (cached, returns deep copy)

**TypeScript:**

- Added `MessageIntrospector<T>` class
- Added `ROSMessageDef` interface

[#1343](https://github.com/RobotWebTools/rclnodejs/issues/1343)

### DIFF
--- a/example/message-introspector/README.md
+++ b/example/message-introspector/README.md
@@ -19,20 +19,12 @@ node message-introspector-example.js
 ### Expected Output
 
 ```
-Type: geometry_msgs/msg/Twist
-Fields: [ 'linear', 'angular' ]
-Defaults: {
-  "linear": {
-    "x": 0,
-    "y": 0,
-    "z": 0
-  },
-  "angular": {
-    "x": 0,
-    "y": 0,
-    "z": 0
-  }
-}
+Twist fields: [ 'linear', 'angular' ]
+Twist defaults: { linear: { x: 0, y: 0, z: 0 }, angular: { x: 0, y: 0, z: 0 } }
+String fields: [ 'data' ]
+String defaults: { data: '' }
+JointState fields: [ 'header', 'name', 'position', 'velocity', 'effort' ]
+Twist schema msgName: Twist
 ```
 
 ## API
--- a/example/message-introspector/README.md
+++ b/example/message-introspector/README.md
@@ -1,0 +1,55 @@
+# ROS 2 MessageIntrospector Example
+
+This directory contains an example demonstrating the `MessageIntrospector` class for inspecting ROS 2 message structure.
+
+## Overview
+
+The `MessageIntrospector` class provides a simple way to understand the structure of ROS 2 messages without directly using `loader.loadInterface`. It's useful for debugging, generating documentation, and building dynamic UIs based on message structure.
+
+## MessageIntrospector Example (`message-introspector-example.js`)
+
+**Purpose**: Demonstrates how to inspect message structure, fields, and default values.
+
+### Run Command
+
+```bash
+node message-introspector-example.js
+```
+
+### Expected Output
+
+```
+Type: geometry_msgs/msg/Twist
+Fields: [ 'linear', 'angular' ]
+Defaults: {
+  "linear": {
+    "x": 0,
+    "y": 0,
+    "z": 0
+  },
+  "angular": {
+    "x": 0,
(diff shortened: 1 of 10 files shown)


example/message-template/message-template-example.js
```diff
@@ -0,0 +1,72 @@
...
+  const fastForward = Twist.clone(forward, { linear: { x: 2.0 } });
+
+  const commands = [
+    { msg: forward, status: 'Moving forward' },
+    { msg: turnLeft, status: 'Turning left' },
+    { msg: backward, status: 'Moving backward' },
+    { msg: turnRight, status: 'Turning right' },
+    { msg: fastForward, status: 'Fast forward!' },
+    { msg: stop, status: 'Stopped' },
+  ];
```
minggangw: Can you please clarify the difference between the implementation here and the original one, like:
```js
  const commands = [
    { msg: { linear: { x: 0.5 }, status: 'Moving forward' },
    { msg: { angular: { z: 1.0 } }, status: 'Turning left' },
    { msg: { linear: { x: -0.5 } }, status: 'Moving backward' },
    { msg: { angular: { z: -1.0 } }, status: 'Turning right' },
  ];

  const { msg, status } = commands[index];
  cmdVelPub.publish(msg);
```

As I understand, the `MessageTemplate` is more likely a `adapter` to get a full-structed object of a specific message. Correct me if I'm wrong, thanks!

minggangw: I want to understand the scenario when developers need `MessageTemplate` and the benefit.

mahmoud-ghalayini: You're correct after testing, I see that `MessageTemplate` doesn't add value over the existing plain object publishing with validation enabled.
My original motivation was that I thought I had to construct the full message object every time, but I now realize rclnodejs already accepts partial objects and fills in defaults automatically. This was a misunderstanding on my side.
I'll close this pull request and the associated issue. However, if you see a way this implementation could be adjusted into something useful for the library, please let me know before I close it.
Thanks for the feedback!

minggangw: One thing I can think about is the `defaults()` method, which can get all fileds defined in a specific msg through a plain object if developers want to understand the structure of it, like
```js
console.log('Twist defaults:', JSON.stringify(Twist.defaults, null, 2));
```

This avoids developers requiring the real message through `loader.loadInterface`.

Others, including `get fields()` and `get schema()`, are useful if developers want to understand the message without touching the complicated one generated by [rosidl_gen](https://github.com/RobotWebTools/rclnodejs/blob/develop/rosidl_gen/index.js)

mahmoud-ghalayini: **Changes I'll make:**

1. **Keep the introspection methods** you mentioned:
   - `defaults` - Get default-initialized plain object
   - `fields` - Get field names
   - `schema` - Get ROSMessageDef schema

2. **Remove message construction utilities** (since plain objects already work):
   - Remove `create()`, `clone()`, `validate()`

3. **Rename to `MessageIntrospector`** (clearer purpose) or keep `MessageTemplate` if you prefer

Does this align with what you had in mind? Happy to adjust based on your preference for naming or API shape

minggangw: Sounds good 👍

example/message-introspector/README.md
```diff
@@ -0,0 +1,55 @@
...
+    "x": 0,
+    "y": 0,
+    "z": 0
+  },
+  "angular": {
+    "x": 0,
+    "y": 0,
+    "z": 0
+  }
+}
```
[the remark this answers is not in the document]
minggangw: We can include the entire result of running the `message-introspector-example.js` here

minggangw (approved):
LGTM, thanks!

PR #1349, opened 2025-12-11
switch to node.js20 package.json

mosfet80:
**Public API Changes**

**Description**


minggangw:
@mosfet80 Can you please give more information about the purpose of this PR, thanks!

mahmoud-ghalayini:
We also might want to add `node-gyp` as a dev dependency

minggangw:
> We also might want to add `node-gyp` as a dev dependency

Yes, we should add `node-gyp` into `devDependencies` explicitly , I will add it, thanks!

Issue #1347, opened 2025-12-09
Add rate limiting options for subscriptions


mahmoud-ghalayini:
### Problem

ROS 2 topics can publish at very high frequencies (e.g., LiDAR at 20Hz, IMU at 100Hz+). When processing these messages in JavaScript/Node.js applications, several problems arise:

1. **Resource exhaustion**: High-frequency callbacks can overwhelm the Node.js event loop, causing UI lag in Electron apps or delayed processing of other events
2. **Unnecessary computation**: Not all use cases require every message (e.g., visualization may only need 5 FPS, logging may only need periodic samples)
3. **Network bandwidth**: When bridging to web clients, transmitting every message wastes bandwidth
4. **No built-in solution**: Unlike ROS 2 C++/Python which can use message filters, rclnodejs lacks native rate limiting support—developers must implement their own throttling logic in every callback

### Solution

Add declarative rate limiting options to `createSubscription()` with three strategies:

- **Throttle**: Limit delivery to a maximum rate (Hz). Useful for capping high-frequency sensors like LiDAR or cameras to a manageable rate for visualization.

- **Downsample**: Deliver every Nth message. Useful for periodic sampling of continuous data streams like IMU or temperature sensors.

- **Debounce**: Wait for a quiet period before delivering the last message. Useful for burst inputs like joystick commands or UI events where only the final settled value matters.

#### API Additions

- `ThrottleStrategy` enum (`DROP`, `LATEST`)
- `throttle`, `downsample`, `debounce` options in `SubscriptionOptions`
- `rateLimitingEnabled` property and `resetRateLimiter()` method on `Subscription`

mahmoud-ghalayini:
## Update: Alternative Approach

I've been working on implementing the rate limiting feature described above, but I realized we might be reinventing the wheel. **What if we use RxJS Observables instead?**

### Current Plan (Custom Rate Limiting)
```javascript
node.createSubscription('sensor_msgs/msg/LaserScan', '/scan', {
  throttle: { maxRateHz: 5 }
}, msg => console.log(msg));
```

### Alternative (RxJS Observables)
```javascript
const { throttleTime, debounceTime, map, filter } = require('rxjs');

node.subscribeAsObservable('sensor_msgs/msg/LaserScan', '/scan')
  .pipe(
    throttleTime(200),  // Same as maxRateHz: 5
    map(msg => msg.data),
    filter(data => data.length > 0)
  )
  .subscribe(data => console.log(data));
[...]

PR #1357, opened 2025-12-17
Add Observable subscriptions with RxJS support

mahmoud-ghalayini:
**Public API Changes**

New method on `Node`:
```typescript
createObservableSubscription<T>(
  typeClass: T,
  topic: string,
  options?: Options,
  eventCallbacks?: (event: object) => void
): ObservableSubscription<MessageType<T>>
```

New exported class `ObservableSubscription`:
| Member | Type | Description |
|--------|------|-------------|
| `observable` | `Observable<T>` | RxJS Observable emitting messages |
| `subscription` | `Subscription` | Underlying ROS 2 subscription |
| `topic` | `string` | Topic name |
| `isDestroyed` | `boolean` | Whether the observable has been completed |
| `complete()` | `void` | Complete the observable |
| `destroy()` | `void` | Alias for `complete()` |

New dependency: `rxjs@^7.8.1`

[#1347](https://github.com/RobotWebTools/rclnodejs/issues/1347)

### DIFF
--- a/example/topics/subscriber/subscription-observable-example.js
+++ b/example/topics/subscriber/subscription-observable-example.js
@@ -39,6 +39,9 @@ async function main() {
     });
 
   // Example 2: Filtered messages (only containing "ROS")
+  // Note: RxJS filter() operates at the application level after messages are received.
+  // For more efficient filtering at the DDS middleware level (reducing network traffic),
+  // use the contentFilter option. See: tutorials/content-filtering-subscription.md
   obsSub.observable
     .pipe(
       map((msg) => msg.data),
--- a/example/topics/subscriber/subscription-observable-example.js
+++ b/example/topics/subscriber/subscription-observable-example.js
@@ -1,0 +1,69 @@
+// Copyright (c) 2025 Mahmoud Alghalayini. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const rclnodejs = require('../../../index.js');
+const { throttleTime, map, filter, bufferCount } = require('rxjs');
+
+async function main() {
+  await rclnodejs.init();
+
+  const node = rclnodejs.createNode('observable_subscription_example_node');
+
+  // Basic observable subscription
+  const obsSub = node.createObservableSubscription(
+    'std_msgs/msg/String',
+    'topic'
+  );
+
+  // Example 1: Throttled messages (max 2/sec)
+  obsSub.observable
+    .pipe(
+      throttleTime(500),
+      map((msg) => msg.data)
+    )
+    .subscribe((data) => {
+      console.log('[Throttled]', data);
+    });
+
+  // Example 2: Filtered messages (only containing "ROS")
+  obsSub.observable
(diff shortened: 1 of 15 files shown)


lib/node.js
```diff
@@ -820,6 +821,82 @@ class Node extends rclnodejs.ShadowNode {
...
+        'eventCallbacks',
+        eventCallbacks,
+        'SubscriptionEventCallbacks',
+        {
+          nodeName: this.name(),
+          entityType: 'subscription',
+          entityName: topic,
+        }
+      );
+    }
```
minggangw: The lines above are dup with
https://github.com/RobotWebTools/rclnodejs/blob/4a1b0db25c4d865bc78bd238bd9d325af6bfcbca/lib/node.js#L762

lib/observable_subscription.js
```diff
@@ -0,0 +1,105 @@
...
+const { Subject } = require('rxjs');
+
+/**
+ * A wrapper that provides RxJS Observable support for ROS 2 subscriptions.
+ * This class wraps a standard Subscription and emits messages through an Observable.
+ *
+ * @class ObservableSubscription
+ * @hideconstructor
+ */
+class ObservableSubscription {
```
minggangw, line 26: I'm think if it's possible to let `ObservableSubscription`  inherit [`Subscription` ](https://github.com/RobotWebTools/rclnodejs/blob/develop/lib/subscription.js)?

example/topics/subscriber/subscription-observable-example.js
```diff
@@ -0,0 +1,69 @@
...
+    )
+    .subscribe((data) => {
+      console.log('[Throttled]', data);
+    });
+
+  // Example 2: Filtered messages (only containing "ROS")
+  obsSub.observable
+    .pipe(
+      map((msg) => msg.data),
+      filter((data) => data.includes('ROS'))
```
minggangw, line 48: As the ROS2 DDS also supports [subscription content filtering](https://docs.ros.org/en/kilted/Tutorials/Demos/Content-Filtering-Subscription.html), which is more efficient I believe, shall we point out the difference between the two `filter`?

minggangw (commented):
Thanks for submitting the PR, the implementation looks elegant! Some comments left, thanks!

lib/observable_subscription.js
mahmoud-ghalayini: Here is the pros, and cons for both implementations.

### Option A: Composition (Current)

`ObservableSubscription` wraps `Subscription`, exposes it via `.subscription`

**Pros:**
- No changes to existing `Subscription` class
- Loose coupling, easier to maintain
- Clear separation: `.observable` for RxJS, `.subscription` for ROS internals

**Cons:**
- Extra wrapper object
- Users access subscription methods via `obsSub.subscription.topic` (though we proxy `topic` directly)

### Option B: Inheritance

`ObservableSubscription extends Subscription`

**Pros:**
- Single object, no wrapper
- Direct access to all `Subscription` methods (`hasContentFilter()`, `publisherCount`, etc.)
- Feels more "native" to the existing API

**Cons:**
- Requires overriding `processResponse()` and duplicating serialization logic
- Tighter coupling to `Subscription` internals
- Need to modify creation in `node.js` to instantiate with native handle directly
- More invasive change

example/topics/subscriber/subscription-observable-example.js
mahmoud-ghalayini: I have updated the readme to have an example for content filter + RxJS

lib/observable_subscription.js
minggangw: I see, let's keep it as a wrapper in order not to couple with current implementation of `Subscription`

minggangw (approved):
LGTM, thanks for your contributions!

minggangw:
@mahmoud-ghalayini I deem the feature implemented by this PR is worth being promoted, can you please add an short description of it into https://github.com/RobotWebTools/rclnodejs/blob/develop/README.md and link it to tutorials/observable-subscriptions.md if the users want to understand more details.

Issue #1458, opened 2026-03-31
Roadmap: Prepare rclnodejs for ROS 2 Lyrical Luth


minggangw:
ROS 2 Lyrical Luth is expected in May 2026 as the next ROS 2 LTS release. For rclnodejs, the main work before release is support enablement rather than major API redesign.

**Lyrical timeline** ([source](https://docs.ros.org/en/rolling/Releases/Release-Lyrical-Luth.html)):
- Alpha + RMW freeze: April 13
- Branch from Rolling: April 21
- Beta: April 27
- GA: May 22

**Default RMW**: `rmw_fastrtps_cpp` (Fast DDS) — unchanged from Jazzy/Kilted.

## Key Focus

- Add `lyrical` to distro detection and internal version mapping.
- Add Lyrical to docs, CI, prebuild workflows, install logic, and native-loader resolution.
- Keep Rolling green and use it as the main pre-release signal for Lyrical readiness.
- Handle the rosidl buffer backend ABI change (ros2/rosidl#941) that affects both Rolling and Lyrical.
- Define a clearer support policy for Humble, Jazzy, Kilted, Lyrical, and Rolling.

### Completed

- [x] Add `lyrical` to distro detection and runtime mapping — #1485 (FUTURE sentinel covers unknown distros)
- [x] Raise the minimum Node.js version to 20.20.2 — #1478
- [x] Add CI coverage for the 2.x minimum Node.js floor — #1478
- [x] Gate C++20 by ROS distro for post-Kilted/Lyrical builds — #1477
- [x] Fix Rolling SIGSEGV caused by rosidl sequence ABI change — #1480
- [x] Remove legacy compatibility code, modernize for Node.js 20+ — #1479
- [x] Treat unrecognized `ROS_DISTRO` as a future distro — #1485
- [x] Reorganize README badges — #1483

### Before Lyrical beta (end of April)

- [ ] Add explicit `LYRICAL` entry to `DistroId` enum and `DistroNameIdMap` in `lib/distro.js`
- [ ] Add Lyrical CI lane (Linux x64) — allow-failure initially
- [ ] Add Lyrical to prebuild generation and artifact naming
- [ ] Update install and native-loader logic for Lyrical binaries
- [ ] Fix conditional include for `rmw/get_service_endpoint_info.h` in `src/rcl_utilities.h` (Rolling/Lyrical removed the transitive include)
[...]

PR #1506, opened 2026-05-07
Run tests on Lyrical instead of Jazzy

minggangw:
Lyrical Luth will be the next ROS 2 LTS, so make it the test target on the Windows pixi job. Flip `run_tests: true` from `jazzy` to `lyrical`; Jazzy, Kilted, and Rolling remain build-only.

Fix: #1458


minggangw:
Lyrical Luth will be the next ROS 2 LTS, so make it the test target on the Windows pixi job. Flip `run_tests: true` from `jazzy` to `lyrical`; Jazzy, Kilted, and Rolling remain build-only.

Fix: #1458

Issue #1594, opened 2026-09-08
Interrupted or failed message generation leaves a partial generated/ tree that init() treats as complete (MESSAGE_NOT_FOUND until generated/ is deleted by hand)


KR-Ravindra:
**Description**

`generateAll()` in `rosidl_gen/index.cjs` writes its version marker (`generated/generator.json`) before it generates a single interface file. If generation does not run to completion, the next `rclnodejs.init()` sees a `generated/` directory whose `generator.json` already carries the current version, decides nothing needs regenerating, and every interface that was never written fails with `MESSAGE_NOT_FOUND`. The library never recovers from this on its own; the only way out is to delete `generated/` manually.

Two realistic ways to end up there:

1. The generating process is killed part-way (SIGKILL/OOM during `npm install`, Ctrl-C on `generate-ros-messages`, container/CI teardown).
2. Any single package fails to generate (EACCES, ENOSPC, a malformed interface). `generateInPath()` runs the packages of a prefix under `Promise.all` (L42-44), so one rejection aborts the loop with the marker already in place. Worse, `scripts/generate_messages.cjs` (the `postinstall` hook and the `generate-ros-messages` bin) catches the error and only logs `Caught error: ...`, so `npm install` exits 0 and the partial tree is left behind silently.

- Library Version: `develop` @ 4bf514e (also 2.2.0, the current release)
- ROS Version: Humble / Jazzy (behaviour is independent of the distro)
- Platform / OS: Linux, Node 18/22

**Where**

`rosidl_gen/index.cjs`, `generateAll()` (L76-107):

- L80-81: `const exist = await fse.exists(generatedRoot); if (forcedGenerating || !exist) {`
- L84: `await fse.emptyDir(generatedRoot);`
- L87-90: `generator.json` is copied into `generated/` -- this is the completion marker, written first
- L94-98: `generated/package.json` and `srv_msg/`
- L103-105: the per-prefix `generateInPath()` loop, i.e. the actual generation, runs after the marker exists

`index.js`, `init()` (L422-431): `forced = version === null || compareVersions(version, generator.version(), '<')` -- the decision to regenerate is based solely on `generated/generator.json`. `getCurrentGeneratorVersion()` (L95-112) returns `null` only when that file is absent. `regenerateAll()` (L558-565) goes through the same `generateAll(true)`.

[...]

Issue #1254, opened 2025-09-07
Interface broken with custom msg: "Error The message required does not exist: {ros2_package}, msg, {ros2_msg} at /{server_dir}/node_modules/rclnodejs/generated/"


VickyXR:
**Description**
I firstly created a ROS2 pkg with **custom msg**, used `colcon build` in this pkg and successfully tested.
Then I would like to get msg from topic defined in this pkg.
When starting the server, I got the error:
```
UNCAUGHT EXCEPTION! System Shutting down...
Error The message required does not exist: {ros2_package}, msg, {ros2_msg} at /{server_dir}/node_modules/rclnodejs/generated/
```
The error still exist even I have sourced the pkg with:
```
source install/setup.bash 
source /opt/ros/humble/setup.bash 
```

My code:

ros2Node.js
```
const rclnodejs = require("rclnodejs");

exports.initROS2 = async () => {
  await rclnodejs.init();
  const node = new rclnodejs.Node("server_node");

  const cmdPublisher = node.createPublisher(
    "pkg/msg/Cmd",
    "/cmd"
  );
  node.subscription = node.createSubscription(
    "pkg/msg/Status",
    "/status",
    (msg) => {
      console.log(
        `[JS]  ${msg.id} status: pos=${msg.position}, err=${msg.error_code}`
      );
    }
  );

  node.spin(node);
  console.log("[ROS2] Node initialized with publisher and subscriber");
  return node;
};
```
server.js
```
const app = require("./app");
const ros2Node = require("./utils/ros2Node");

// start ROS2 node
const node = ros2Node.initROS2();

// start server
const port = 3000;
app.listen(port, () => {
  console.log(`App running on port ${port}...`);
});
```
I have tried to make my server project as a ros2 pkg, 
`ros2 pkg create web_interface --build-type ament_nodejs --dependencies rclnodejs`
but got error:
`ros2 pkg create: error: argument --build-type: invalid choice: 'ament_nodejs' (choose from 'cmake', 'ament_cmake', 'ament_python')`

using npx `npx generate-ros-messages`
got error: `sh: 1: generate-ros-messages: Exec format error`

using npm `npm run generate-messages-idl`
got error: 
```
npm error Missing script: "generate-messages-idl"
npm error
npm error To see a list of scripts, run:
npm error   npm run
npm error A complete log of this run can be found in: /home/vicky/.npm/_logs/2025-09-07T06_21_38_347Z-debug-0.log
```

[...]

Issue #688, opened 2020-07-30
Types for creating an Action Client not Updating with Sources Action Message Types


caelinsutch:
Problem example:

```js
// Variable actionType has to be a certain string value. Types don't update when sourcing additional packages
const newActionClient = new ActionClient(node, actionType as ActionTypeClassName, actionName);

newActionClient.sendGoal(data).then(v => {
  log.debug('Action Complete!');
})
```

mattrichard:
Right now the message types are only generated during postinstall or at runtime when calling `init()`. I can see how this would be a problem with TypeScript since you need to be able to compile everything before `init()` can be run again. We'll either need to make the action types more flexible or find better ways to update the generated message typings when additional packages are sourced.

For now, you can workaround this issue by manually regenerating the message typings using the following command. This should at least unblock you until there's a proper solution.

```bash
node node_modules/rclnodejs/scripts/generate_messages.js
```

minggangw:
We once planed to implement that the message generator could generate the JS messages from IDL dynamically if the required message cannot be found, but can be found through `PATH` (the new package will be added after you source the `xxx_install.sh`). 

Unfortunately, you have to manually delete the `generated\` folder in order to trigger to generate the messages completely again, or workaround as @mattrichard said. Thanks!

caelinsutch:
> For now, you can workaround this issue by manually regenerating the message typings using the following command. This should at least unblock you until there's a proper solution.

I got this error:
```
Start JavaScript message generation...
Caught error: TypeError: Cannot read property 'fields' of undefined
```

caelinsutch:
Update - Even after deleting `generated` and even the entire folder and reinstalling, I'm still not getting all my packages sourced. For example, when calling a the service `/save_maps` from Slamtoolbox, I'm getting the error `The message required does not exist: slam_toolbox, srv, SaveMap`. In the same terminal window, I can run that service from the command line. 

This is from the CLI: 

```

➜ ros2 service type /save_map 
slam_toolbox/srv/SaveMap

```

wayneparrott:
Just saw your updates. Thx for sharing.
I installed slam_toolbox and believe the error message you shared is from the generate_tsd.js script resulting in the interfaces.d.ts file having incomplete slam_toolbox message types defined. Will dig into this asap.
```
Start JavaScript message generation...
Caught error: TypeError: Cannot read property 'fields' of undefined
```
[...]

Issue #546, opened 2019-12-13
Incomplete installation process on v0.11.0 - Message generation failing


wayneparrott:
I while installing rclnodejs v0.11.0 from npm the message generation process is aborting when it encounters a missing file. This results in only a subset of js file being generated and the interface.d.ts is not generated. Here's the informational portion of stdout during the installation process.  

Note: the same issue applies when running `rclnodejs#regenerateAll()`

We need to fix this immediate issue and post a 0.11.1 update as well as extend testing to detect message generation issues.

```
> rclnodejs@0.11.0 postinstall /home/wayne/dev/workspaces/ros-lidar/test1/node_modules/rclnodejs
> node scripts/generate_messages.js

Start JavaScript message generation...
Caught error: Error: Cannot find module '../../generated/action_tutorials_interfaces/action_tutorials_interfaces__msg__FibonacciActionGoal.js'
npm WARN typescript-hello-world@1.0.0 No repository field.

+ rclnodejs@0.11.0
added 55 packages from 36 contributors and audited 77 packages in 51.97s
found 0 vulnerabilities
```

wayneparrott:
How come we didn't find this bug when testing?

I checked the CI logs and for example the travis ci is using the same Linux version as I am but it is not reproducing this problem when it runs the installation step. And in my dev environment I had an older copy of the generated message files prior to our removal of processing action files. So my local env did not reflect a proper user environment. Once I removed the generated folder from my dev environment I am able to replicate the failure.

wayneparrott:
Added try/catch around tsd message file process to ignore errors as it is a non-critical failure. The output will appear as follows:

```
Start JavaScript message generation...
Unable to generate tsd info: Cannot find module '../../generated/action_tutorials_interfaces/action_tutorials_interfaces__msg__FibonacciActionGoa
l.js'
Unable to generate tsd info: Cannot find module '../../generated/action_tutorials_interfaces/action_tutorials_interfaces__msg__FibonacciFeedback.
js'
Unable to generate tsd info: Cannot find module '../../generated/action_tutorials_interfaces/action_tutorials_interfaces__msg__FibonacciGoal.js'
Unable to generate tsd info: Cannot find module '../../generated/action_tutorials_interfaces/action_tutorials_interfaces__msg__FibonacciResult.js
'
Unable to generate tsd info: Cannot find module '../../generated/example_interfaces/example_interfaces__msg__FibonacciActionGoal.js'
[...]

PR #1595, opened 2026-09-08
Write generated/generator.json only after message generation completes

KR-Ravindra:
## Problem

If message generation does not run to completion, `rclnodejs.init()` keeps treating the partial `generated/` tree as complete. Every interface that was never written then fails with `MESSAGE_NOT_FOUND` on every start, and the only way out is to delete `generated/` by hand. The two realistic ways to get there are the generating process being killed part-way (SIGKILL/OOM during `npm install`, Ctrl-C on `generate-ros-messages`, CI/container teardown) and any single package failing to generate (EACCES, ENOSPC, a malformed interface). In the second case `npm install` also exits 0, because the `postinstall` script only logs the error, so the partial tree is left behind silently.

## Root cause

- `rosidl_gen/index.cjs:87-90` copies `generator.json` (the version marker `init()` uses to decide whether regeneration is needed) into `generated/` before the per-prefix `generateInPath()` loop at `rosidl_gen/index.cjs:103-105` has generated a single interface file.
- `index.js:422-424` (`getCurrentGeneratorVersion()` / `forced`) only regenerates when that file is absent or carries an older version, so a partial tree with a current-version marker is never regenerated.
- `scripts/generate_messages.cjs:36-38` catches the error, prints `Caught error: ...` and returns normally, so the process exits 0.

## Fix

- `rosidl_gen/index.cjs`: write `generated/generator.json` last, after the `generateInPath()` loop. `generated/` is now created with `fse.mkdirs()` up front (the `generator.json` copy used to create it implicitly). A tree left behind by an interrupted or failed run has no marker, `getCurrentGeneratorVersion()` returns `null`, and `init()` regenerates it on the next start with no other change.
- `scripts/generate_messages.cjs`: `process.exit(1)` in the catch handler so `npm install` and `npx generate-ros-messages` report the failure instead of succeeding with a partial tree.
- `test/test-messsage-generation-overlay.js`: regression test that points `AMENT_PREFIX_PATH` at a prefix whose ament index lists a `.msg` file that does not exist, runs `generateAll(true)`, and asserts that it fails and that no `generated/generator.json` was written.

## How tested

New test against the unmodified generator:

```
  override interface in overlay tests
    1) failed generation leaves no generator.json marker
  0 passing (100ms)
  1 failing
  1) override interface in overlay tests
       failed generation leaves no generator.json marker:
      AssertionError [ERR_ASSERTION]: Expected no generator.json after a failed generation
      + expected - actual
      -false
      +true
```

With the fix:

```
  override interface in overlay tests
    ✔ failed generation leaves no generator.json marker (148ms)
  1 passing (156ms)
```

Also checked that a successful `generateAll(false)` followed by `generateAll(true)` still produces `generated/generator.json`, `generated/package.json` and `generated/srv_msg/`, and that `node scripts/generate_messages.cjs` now exits 1 when a package fails to generate (`generated/` is left without `generator.json`). `npx prettier --check` and `npx eslint` pass on the changed files.

## Links

- https://github.com/RobotWebTools/rclnodejs/issues/1594
- Earlier reports of users stuck in this state: https://github.com/RobotWebTools/rclnodejs/issues/546, https://github.com/RobotWebTools/rclnodejs/issues/688, https://github.com/RobotWebTools/rclnodejs/issues/1254

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra (commented):
Round 1 self-review.

Checked against `develop`: `rosidl_gen/index.cjs` copies `generator.json` into `generated/` before the `generateInPath()` loop, `init()` only regenerates when that marker is missing or older than the generator, and `scripts/generate_messages.cjs` logs the error and returns normally, so the description matches the code.

1. Marker is now written after the loop and `generated/` is created with `fse.mkdirs()` up front (the copy used to create it implicitly), so an interrupted or failed run leaves no marker and the next `init()` regenerates. Looks correct.
2. `process.exit(1)` in the postinstall catch makes `npm install` report the failure instead of exiting 0 with a partial tree.
3. The new mocha case uses the file's existing `generateMessages()` / `GENERATED_PATH` helpers, restores `AMENT_PREFIX_PATH` and removes the temp prefix in `finally`.

CI: full ROS 2 matrix (humble, jazzy, kilted, lyrical, rolling; arm64 + x64; pixi builds) green, coverage unchanged. Marking ready.

minggangw:
lgtm, merging

minggangw (approved):

Issue #1330, opened 2025-12-01
Missing rcl functions in rclnodejs


minggangw:
### Missing `rcl` Functions in `rclnodejs`

**Action**
*   `rcl_action_goal_status_array_fini` #1332

**Arguments**
*   `rcl_arguments_get_unparsed_ros` #1342
*   `rcl_remove_ros_arguments` #1339

**Callbacks & Events**
*   `rcl_client_set_on_new_response_callback`
*   `rcl_event_set_callback`
*   `rcl_service_set_on_new_request_callback`
*   `rcl_subscription_set_on_new_message_callback`

**Clock & Time**
*   `rcl_clock_add_jump_callback` #1355
*   `rcl_clock_remove_jump_callback` #1355
*   `rcl_difference_times` #1354 
*   `rcl_timer_clock`
*   `rcl_timer_get_next_call_time` #1356
*   `rcl_timer_set_on_reset_callback` #1348

**Introspection (Service & Graph)**
*   `rcl_get_clients_info_by_service` #1335
*   `rcl_get_info_by_service` #1335
*   `rcl_get_servers_info_by_service` #1335
*   `rcl_get_zero_initialized_service_endpoint_info_array` #1335
*   `rcl_service_endpoint_info_array_fini` #1335

**Logging & Rosout**
*   `rcl_logging_configure_with_output_handler` #1338
*   `rcl_logging_get_logging_directory` #1337
*   `rcl_logging_multiple_output_handler` #1338
*   `rcl_logging_rosout_add_sublogger` #1336
*   `rcl_logging_rosout_enabled` #1336
*   `rcl_logging_rosout_fini_publisher_for_node` #1336
*   `rcl_logging_rosout_init_publisher_for_node` #1336
*   `rcl_logging_rosout_remove_sublogger` #1336

**Publisher**
*   `rcl_publisher_assert_liveliness` #1331

**Subscription (Content Filtering)**
*   `rcl_subscription_content_filter_options_fini` #1334
*   `rcl_subscription_get_content_filter` #1334

**Service**
*   `rcl_take_request_with_info` #1333

PR #1599, opened 2026-09-14
Add Timer.clock accessor

KR-Ravindra:
Adds the `rcl_timer_clock` item of the parity checklist in #1330: a `Timer.clock` accessor, matching rclpy's `Timer.clock`.

## Design

- `timer.clock` returns the `Clock` the timer was created with: the clock passed to `node.createTimer()`, or the node's clock when none was passed. `Node.createTimer()` now hands that clock to the `Timer` constructor. No new native binding: the addon is unchanged.
- Behaviour change: Because a `Timer` now keeps a reference to its `Clock`, `Node.destroy()` releases the timer handles explicitly. Timer handles are children of the clock handle rather than of the node handle, so until now they were finalized only once the clock was garbage collected; that also kept the thread-safe function of `setOnResetCallback()` alive.
- `types/timer.d.ts` gets `readonly clock: Clock` and the tsd test covers it.

## How tested

`mocha test/test-timer.js` on ROS 2 Kilted (Node 22), with the new tests applied before the change:

```
  24 passing (2s)
  2 failing

  1) timer.clock returns the clock the timer was created with:
     AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
     + undefined
     - ROSClock { _clockType: 1, _handle: RclHandle {} }
  2) timer.clock exposes ROS time of a ROSClock timer: (same assertion)
```

After the change: `26 passing, 1 pending`, and the process exits on its own. The new tests check the node clock, a custom `STEADY_TIME` clock passed as the 4th argument, a `ROSClock` passed as the 3rd argument with a ROS time override read back through `timer.clock.now()`, and that `timer.clock` is read-only. The full suite in the same container: 1454 passing, 14 pending; the 2 failures are HTTP-transport timeouts in `test/test-web-http.js` that fail identically on unmodified `develop` there. `prettier`, `eslint` and `tsd` are clean.

Revision 2 (review): the `getTimerClock` native binding and the tests that compared its result with `clock.handle` were removed; the remaining tests are unchanged.

Revision 3 (review): dropped the `clock = null` constructor default and added a CONTRIBUTORS.md entry; no functional change.

## Links

- https://github.com/RobotWebTools/rclnodejs/issues/1330 (`rcl_timer_clock` under "Clock & Time")
- rclpy `Timer.clock`: https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/timer.py

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.
1. `CreateTimer` already parents the timer handle to the clock handle (`src/rcl_timer_bindings.cpp:131`), so `getTimerClock` returning the parent's JS object avoids wrapping the same `rcl_clock_t` twice.
2. `rcl_timer_clock` is declared in rcl's `timer.h` on humble through rolling, so the binding needs no `ROS_VERSION` guard.
3. The node clock is created with no parent handle, which is why `node.destroy()` never released timer handles before; `RclHandle::Release` is idempotent, so the explicit release is safe next to `destroyTimer()`.
4. The one red job (rolling on `ubuntu:resolute`) died in "Enable ROS2 apt repository" before the build step; unrelated to this diff.
Leaving as draft until the rest of the matrix finishes.

KR-Ravindra:
Round 2 self-review.

CI: 15 of 16 jobs pass. The one red job (build-and-test 26.X x64 rolling, ubuntu:resolute) fails in the Node setup step before checkout: `node: error while loading shared libraries: libatomic.so.1`. The same job with the same Node 26.8.2 passed on develop on 2026-09-11 (run 34552620758), so this looks like the freshly pulled resolute image, not this change. I cannot rerun it from a fork; a maintainer rerun should clear it.

Marking ready: change verified against develop as in Round 1, all other Linux/arm64/Windows/ASan/pixi jobs green, coveralls pass.

src/rcl_timer_bindings.cpp
```diff
@@ -291,6 +291,29 @@ Napi::Value GetTimerPeriod(const Napi::CallbackInfo& info) {
...
+  RclHandle* clock_handle = timer_handle->parent();
+  if (clock_handle == nullptr || clock_handle->ptr() != clock) {
+    Napi::Error::New(env, "Timer clock handle is not available")
+        .ThrowAsJavaScriptException();
+    return env.Undefined();
+  }
+
+  return clock_handle->Value();
+}
+
```
minggangw: Hi @KR-Ravindra thanks for submitting the PR, I prefer not exposing this C++ binding method, because it will not be used. Instead, developers can call the `get clock()` added into `Timer` class without an additional native lookup.

KR-Ravindra: @minggangw done in c93eaf2: dropped the `getTimerClock` binding (src/ is now untouched) and the three test assertions that compared its result with `clock.handle`; `Timer.clock` stays a JS-only getter fed by `Node.createTimer()`. Could you take another look?

lib/timer.js
```diff
@@ -21,9 +21,10 @@ import DistroUtils from './distro.js';
  */
 
 class Timer {
-  constructor(handle, period, callback) {
+  constructor(handle, period, callback, clock = null) {
```
minggangw: nit: I think setting a default value of `null` for `clock` isn't need here, btw, pls feel free to add yourself as contributor into https://github.com/RobotWebTools/rclnodejs/blob/develop/CONTRIBUTORS.md

LGTM, thanks!

minggangw (approved):
Pls squash merge the PR when you address the comments, thanks!

lib/timer.js
KR-Ravindra: @minggangw done in 567cf1b: dropped the `= null` default (the constructor is only called from `Node.createTimer()`, which always passes a clock) and added myself to CONTRIBUTORS.md, thanks. prettier and eslint are clean. I don't have write access to the repo, so I can't squash-merge it myself; could you merge it when convenient?

minggangw: Merging...

